PR #1, opened 2018-05-31
Added keypad library

sushshring:


mcderdav:
To my mental compiler, it should work aside the improper constructor call marked as requested changes.  If it passes verification in Arduino studio, my comments can be disregarded.  Otherwise, looks great!

sushshring:
See [Arduino Playground](http://playground.arduino.cc/Code/Keypad#Example). Code referenced from here

mcderdav:
That's super weird and shouldn't work, but if it does, it does, so I will yield to the playground.

PR #2, opened 2018-06-07
Fixed compile issue with RFID

mcderdav:


sketch/sketch.ino
```diff
@@ -6,11 +6,11 @@ Scanner scanner(0,0);
...
-  
+  Serial.begin(9600);
 }
 
 void loop() {
-
+  Serial.println(scanner.getUID());
 }
 
 void chageState(){
```
sushshring: nit: Spelling error

sketch/hackPSUrfid.cpp
```diff
@@ -2,6 +2,8 @@
 
 namespace hackPSU {
 
+constexpr byte Scanner::DEFAULT_KEY[6];
```
sushshring: Is this the key to send to Redis?

mcderdav: No, this is the auth key for the wristbands.  The library provides a few defaults in a text file, but this is the key we used last time

PR #4, opened 2018-06-14
Adding LCD Functions.

ThePBJain:


kramersd:
Looks good to me.

mdh5389:
"Loc: " + location + ", " + wifi + "%" // location is a string and wifi is an int (Line 52)
This string concatenation is not valid.  Other than that, Looks good to me.

Also, not sure if we want to wrap the liquidcrystal class so we can do, for example, lcd.printName(...)

HackPSUlcd/HackPSUlcd.ino
```diff
@@ -25,9 +25,67 @@ void setup() {
 
   // Move the cursor to the next line and print
   // WORLD.
-  lcd.setCursor(5, 1);      
+  lcd.setCursor(5, 1);
   lcd.print("WORLD");
 }
 
 void loop() {
+  printDebug("Food", 88);
```
sushshring: Is the loop code only for testing?

HackPSUlcd/HackPSUlcd.ino
```diff
@@ -25,9 +25,67 @@ void setup() {
...
+}
+
+void printRegistered(String registered){
+  lcd.clear();
+  lcd.setCursor(0, 1);
+  lcd.print("Registered: " + registered);
+}
+
+
+void printMsg(String msg){
```
sushshring: Since `printMsg()` is the generic version for all these other methods, is there merit in making them call printMsg?

HackPSUlcd/HackPSUlcd.ino
ThePBJain: yea

HackPSUlcd/HackPSUlcd.ino
```diff
@@ -25,9 +25,67 @@ void setup() {
...
+}
+
+
+void printSize(String size){
+    lcd.clear();
+    lcd.setCursor(0, 1);
+    lcd.print("Size: " + size);
+}
+
+void printScan(bool goodScan){
```
sushshring: nit: goodScan is a little confusing as a variable name. Can we make is `isGoodScan` or something, so that its more obvious that it can take both values?

HackPSUlcd/HackPSUlcd.ino
sushshring: Okay cool

HackPSUlcd/HackPSUlcd.ino
ThePBJain: Rewrote methods from Python class. I think we want this one if we need a message that is non traditional.

PR #5, opened 2018-07-09
Add more functions to HTTP wrapper

jmm8046:
what even is github did i do this right?


mcderdav:
I just fixed your merge conflicts, if Sush hasn't already explained what they are to you, raise that in the next meeting and one of us will explain it so that you can handle those yourself.  Now for review time

sketch/hackPSUhttp.cpp
```diff
@@ -1,25 +1,72 @@
...
-		    response.responseCode = httpCode;
-	    } else {
-		    response.payload = http.errorToString(httpCode).c_str();
-		    response.responseCode = httpCode;
-	    }
-	http.end();
-	return responseInfo;
-	};
-}
+  	Response::GET(String url, String payload, struct headers){
```
mcderdav: Quick note, in your header, you did 
```C
struct Response {
    String payload;
    int responseCode;
} responseInfo;
```

This creates a struct that calls itself Response, but that everyone else has to call responseInfo.  This will not work, and here, you will not have a name accessible as a return type.  Your header should have something like:
```C
typedef struct Response {
    ...
} Response;
```
It looks redundant, but the firs instance of `Response` names the struct as `struct Response` and the later one names `struct Response` `Response`

sketch/hackPSUhttp.cpp
```diff
@@ -1,25 +1,72 @@
...
+  		HTTPClient http;
+   		http.begin(url); //http Begin call
+    		int httpCode = http.GET(); //GET call
 
+    		http.addHeader(headerInfo.header);
```
mcderdav: Where is headerInfo defined?

sketch/hackPSUhttp.cpp
```diff
@@ -1,25 +1,72 @@
...
 
+  		//httpCode will be negative (or zero) if there is an error
+  		if (httpCode > 0){ 	//If the GET call was successful
+      			responseInfo.payload = http.getString(); 		//getting the repsonse from the get
+      			responseInfo.responseCode = httpCode;
+  		}else{
+     			responseInfo.payload = http.errorToString(httpCode).c_str(); //I think this is the error information??
+     			responseInfo.responseCode = httpCode; //should be negative or zero?
+  			}
```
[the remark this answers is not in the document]
mcderdav: } should be one tab in

sketch/hackPSUhttp.cpp
```diff
@@ -1,25 +1,72 @@
...
+      			responseInfo.payload = http.errorToString(httpCode).c_str(); //I think this is the error information??
+      			responseInfo.responseCode = httpCode; //should be negative or zero?
+  		}
+    		http.end();
+  	}
+
+  	Response::POST(String url, String payload){
+    		HTTPClient http;
+    		http.begin(url); //http Begin call
+    		int httpCode = http.GET(); //GET call
```
mcderdav: Why are we performing a get call when the function name indicates a post call?

sketch/hackPSUhttp.cpp
```diff
@@ -1,25 +1,72 @@
...
+    		http.begin(url); //http Begin call
+   		int httpCode = http.GET(); //GET call
+    		http.addHeader(headerInfo.header);
 
+  		//httpCode will be negative (or zero) if there is an error
+  		if (httpCode > 0){ 	//If the GET call was successful
+     			responseInfo.payload = http.getString(); 		//getting the repsonse from the get
+      			responseInfo.responseCode = httpCode;
+  		}else{
+      			responseInfo.payload = http.errorToString(httpCode).c_str(); //I think this is the error information??
```
mcderdav: We probably don't want c_str() anywhere; Arduino likes String, which is nice and abstract, c_str is literally an array of chars

sketch/hackPSUhttp.cpp
```diff
@@ -1,25 +1,72 @@
...
-}
+  	Response::GET(String url, String payload, struct headers){
+  		HTTPClient http;
+   		http.begin(url); //http Begin call
+    		int httpCode = http.GET(); //GET call
 
+    		http.addHeader(headerInfo.header);
 
+  		//httpCode will be negative (or zero) if there is an error
+  		if (httpCode > 0){ 	//If the GET call was successful
```
mcderdav: Is httpCode the http status code or is that an error code out of the function.  If it's the request status code, 200 is OK, most others are errors.  If it's a status code, you can simplify the condition to `if (httpCode)`

sketch/hackPSUhttp.cpp
sushshring: Just merged that in. May need renaming, but its there now.

sushshring:
@jmm8046 Can we fix up the changes @mcderdav requested, merge this branch into patch-1, and review all changes based on the difference from master?
P.S. You can just continue pushing commits to this branch. No need to PR each time :)

sushshring:
Looking much better! I think the only thing that hasn't been addressed was the `struct responseInfo`

sketch/hackPSUrfid.cpp
```diff
@@ -1,53 +1,74 @@
-#include "hackPSUrfid.h"
+//https://techtutorialsx.com/2016/07/21/esp8266-post-requests/
+#include <hackPSUhttp.h>
```
sushshring: Why is this implementation in hackPSUrfid.cpp? Shouldn't this be in hackPSUHTTP.cpp?

sushshring:
Looks good to me!

sketch/hackPSUhttp.cpp
```diff
@@ -1,25 +1,74 @@
+//https://techtutorialsx.com/2016/07/21/esp8266-post-requests/
 #include <hackPSUhttp.h>
```
mcderdav: Local headers use `"` quotes not `<` brackets

sketch/hackPSUhttp.h
```diff
@@ -5,25 +5,25 @@
...
 
   class HTTP{
     public:
-	  /**
-	   * TODO: Add docstring explaining these parameters
-	   */
+    /**
+     * TODO: Add docstring explaining these parameters
+     */
       static Response GET(String url, String headers)
```
mcderdav: Semicolons bruh

sketch/hackPSUhttp.h
```diff
@@ -5,25 +5,25 @@
...
       static Response GET(String url)
       static Response POST(String url, String payload, String headers)
       static Response GET(String url, String payload)
```
mcderdav: CPP functions names are mangled as `_NAME_ARG-TYPE-1_ARGE-TYPE-2...` you cannot overload with two different strings; at least one must have a different type or be removed

sketch/hackPSUhttp.cpp
```diff
@@ -1,25 +1,74 @@
...
-		    response.responseCode = httpCode;
-	    } else {
-		    response.payload = http.errorToString(httpCode).c_str();
-		    response.responseCode = httpCode;
-	    }
-	http.end();
-	return responseInfo;
-	};
-}
+	Response::GET(String url, String payload, struct headers){
```
mcderdav: This function needs a return type

mcderdav: also others need that

sketch/hackPSUhttp.h
mcderdav: These function bois need return types (probably `void`) use `void` for no return, or `unsigned short short` for status codes

sketch/hackPSUhttp.cpp
```diff
@@ -1,25 +1,74 @@
...
+		HTTPClient http;
+   		http.begin(url); //http Begin call
+    		http.addHeader(Headers.headerInformation);
+		int httpCode = http.GET(); //GET call
 
 
+  		//httpCode will be negative (or zero) if there is an error
 
+  		//HAVE IF FOR IF NEGATIVE AND ERROR CONNECTION TO WIFI??
+  		if (httpCode > 0){ 	//If the GET call was successful
```
ThePBJain: Where are these codes coming from?

sketch/hackPSUhttp.cpp
```diff
@@ -1,25 +1,74 @@
...
+      			responseInfo.payload = http.getString(); 		//getting the repsonse from the get
+      			responseInfo.responseCode = httpCode;
+  		}else{
+     			responseInfo.payload = http.errorToString(httpCode); //I think this is the error information??
```
kramersd: The error message on a response (if properly handled by the server) will be sent as a payload. The httpCode is just a number. Instead of setting the payload of the response back as the httpCode, it should be changed to the payload (usually a JSON message) which will state the issue as a string. Otherwise you will have the httpCode being an error (401) and the payload being an error (401). But the message (unauthorized) will never be displayed.

kramersd: Also change this across the other functions that do the same thing.

sketch/hackPSUhttp.cpp
```diff
@@ -1,25 +1,74 @@
...
-	    } else {
-		    response.payload = http.errorToString(httpCode).c_str();
-		    response.responseCode = httpCode;
-	    }
-	http.end();
-	return responseInfo;
-	};
-}
+	Response::GET(String url, String payload, struct headers){
+		HTTPClient http;
```
sushshring: Should these be malloced using new rather than be on the stack?

sushshring: Is yes calls will have to change to `HTTPClient* http = new HTTPClient()`, and method calls will be `http->begin()`

sketch/hackPSUhttp.cpp
sushshring: These httpCodes are part of the underlying HTTP library. Positive codes mean successful requests, and the response will have a separate code for the actual HTTP repose.

sketch/hackPSUhttp.cpp
```diff
@@ -1,25 +1,74 @@
...
+  	}
 
+  	Response::GET(String url, String payload){
+    		HTTPClient http;
+    		http.begin(url); //http Begin call
+   		int httpCode = http.GET(); //GET call
+
+  		//httpCode will be negative (or zero) if there is an error
+  		if (httpCode > 0){ 	//If the GET call was successful
+     			responseInfo.payload = http.getString(); 		//getting the repsonse from the get
```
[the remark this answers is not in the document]
sushshring: I think what is needed here is a new allocation for a responseInfo object. Currently I think the global definition from the header file is being used, which probably won't work.

sketch/hackPSUhttp.h
```diff
@@ -5,25 +5,25 @@
...
-  		int responseCode;
-	}responseInfo;   //declare structure obejct to return
+  
+  typedef struct Response {
+      String payload;
+      int responseCode;
+  }Response;   //declare structure obejct to return
+
+  typedef struct Headers{   //does this belong here??
+       String headerInformation;
```
sushshring: Since these headers are key/value pairs, I think it might be more beneficial to break them apart into String headerKey, String headerValue. Then when adding the header, you can use a helper method that will convert them to a string representation like `Content-Type: application/json`.

sketch/hackPSUhttp.h
sushshring: Rather than String headers then, these functions could take a variable amount of headers `static Response POST(String url, String payload, Headers... headers)`, which basically translates to `Headers[]`.

sketch/hackPSUhttp.h
sushshring: Change to static Response POST(String url, String payload)

sketch/hackPSUhttp.h
sushshring: The return type is Response @mcderdav

sketch/hackPSUhttp.h
jmm8046: I'm not sure what you mean by this. Should I change it to this or do you mean something else (also is this where you wanted the semicolons in the other comment?)
```
static Response GET(String url, headers);
static Response GET(String url);
```

sketch/hackPSUhttp.cpp
jmm8046: I'm not sure what you mean by malloced so I have no idea if I should change this or not

sketch/hackPSUhttp.h
mcderdav: @jmm8046 , Yeah, that works, you just can't have two functions with the same argument types.  Lines 23 and 26 conflicted in this example.  As for semicolons, almost every  line needs to end with a semicolon, the only exception being after a code block (except blocks that define structures).  So all your prototypes need `;` at the end.  You're proposed revision above is perfectly valid C++.  The only thing you may want to consider is that in C/C++ names in ALL CAPS are generally definitions, so you should avoid all caps names for functions and variables, as that is reserved for `#define`, `constexpr`, & `const` statements

sketch/hackPSUhttp.h
mcderdav: Shit I got too aggressive, u rite

sketch/hackPSUhttp.cpp
mcderdav: One malloc boi gives you space to store variables, so Sush is saying that you'll allocate memory for it so that other functions can access it.  You technically can keep it on the stack if you want, but it may make your life more difficult than it's worth

PR #8, opened 2018-08-29
Added HTTP Wrapper

sushshring:


httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,91 @@
...
+		if (response->responseCode < 0){
+			throw "http request failed :(";
+		}
+
+		StaticJsonBuffer<200> jsonBuffer;
+		JsonObject& root = jsonBuffer.parseObject(response->payload);
+
+		if (!root.success()) {
+			throw "json parsing failed :( lit";
+  		}
```
sushshring: Since we don't know what the schema of the returned object will be quite just yet, can we add a comment here that says `// Parse from JSONObject into type jawn`

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+	static jawn assignRfidToUser(String userId, String userBandId){ //userId sent in from band, locationId sent in from device
+		/*
+		post request sent to redis with userId and locationId     redis url: tabs/setup
+		redis returns if they are good or not
+		return ok or not
+		*/
+
+		String url = "https://"+redisHost+"/tabs/setup";
+		String payload = "{\"userID\":"+userId+",\"locationId\":"+locationId+"}";
```
sushshring: userBandId is never used.

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,91 @@
...
+		if (response->responseCode < 0){
+			throw "http request failed :(";
+		}
+
+		StaticJsonBuffer<200> jsonBuffer;
+		JsonObject& root = jsonBuffer.parseObject(response->payload);
+
+		if (!root.success()) {
+			throw "json parsing failed :(";
+  		}
```
sushshring: Parse result from `root` and return a boolean. Currently this won't compile because no value is returned.

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	static bool entryScan(String userBandId, String locationId){
+		/*
+		post request sent to redis with userId and locationId      redis url: tabs/getpin
+		parse through above information and assign to variable(s)
+		return variables (whether they are good to go or not)
+		*/
+
+		String url = "https://"+redisHost+"/tabs/getpin";
```
sushshring: This seems the wrong url?

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+		StaticJsonBuffer<200> jsonBuffer;
+		JsonObject& root = jsonBuffer.parseObject(response->payload);
+
+		if (!root.success()) {
+			throw "json parsing failed :(";
+  		}
+
+  		//The following is based on assumptions and should be checked
+  		bool entryAllowedOrNot = root["data"][0];
```
sushshring: nitpick: Redundant variable assignment. Could be returned directly as long as you cast it as either `root["data"][0].asBool()` (Check if this library supports that), else simply `(bool) root["data"][0]`

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+		StaticJsonBuffer<200> jsonBuffer;
+		JsonObject& root = jsonBuffer.parseObject(response->payload);
+
+		if (!root.success()) {
+			throw "json parsing failed :( lit";
+  		}
+	}
+
+	static jawn assignRfidToUser(String userId, String userBandId){ //userId sent in from band, locationId sent in from device
```
sushshring: Since these are class members, they should be `static jawn HTTPImpl::assignRfidToUser(...) {...}`?

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,91 @@
+#include "httpwrapper.h"
+
+
+namespace hackPSU {
+	static jawn getDataFromPin(int pin){	//person inserts pin
```
sushshring: Now that we have the actual definition of the data the scanner receives from Redis-Server, redefine `jawn` to be an actual struct and parse the data coming in into the struct.

sushshring:
Could we rename `jawn` to something like `redisData`?

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,119 @@
...
+		Response* response = HTTP.POST(url, payload, count, headers);
+
+		if (response->responseCode < 0){
+			throw "http request failed :(";
+		}
+
+		StaticJsonBuffer<200> jsonBuffer;
+		JsonObject& root = jsonBuffer.parseObject(response->payload);
+
+		String userUid =root["uid"];
```
sushshring: All of these will require `asString()` called on them.

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,119 @@
...
+		String userShirtSize =root["shirtSize"];
+		String userDiet =root["diet"];
+		String userCounter =root["counter"];
+		String userNumScans =root["numScans"];
+
+		/*I was not sure if if could have
+			pinData->uid = uid 
+			and have the variables that same name so I tried to change them
+		*/
+		jawn* pinData;
```
sushshring: Combine the declaration and instantiation? `jawn* pinData = new jawn`?

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,119 @@
...
+		String userPin =root["pin"];
+		String userName =root["name"];
+		String userShirtSize =root["shirtSize"];
+		String userDiet =root["diet"];
+		String userCounter =root["counter"];
+		String userNumScans =root["numScans"];
+
+		/*I was not sure if if could have
+			pinData->uid = uid 
+			and have the variables that same name so I tried to change them
```
sushshring: You can have them be the same name, since they're part of the struct.

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,119 @@
...
+		*/
+		jawn* pinData;
+		pinData = new jawn;
+		pinData->uid = userUid;
```
sushshring: Could these be read directly from the JsonObject? Would reduce code lines.

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,119 @@
...
+		pinData->givenPin = userPin;
+		pinData->name = userName;
+		pinData->shirtSize = userShirtSize;
+		pinData->diet = userDiet;
+		pinData->counter = userCounter;
+		pinData->numScans = userNumScans;
+
+		// Parse from JSONObject into type jawn
+		if (!root.success()) {
+			throw "json parsing failed :( lit";
```
sushshring: <3

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,119 @@
...
+		JsonObject& root = jsonBuffer.parseObject(response->payload);
+		
+		// Parse from JSONObject into type jawn
+		if (!root.success()) {
+			throw "json parsing failed :(";
+  		}
+
+  		//The following is based on assumptions and should be checked
+
+  		return (bool) root["data"][0];
```
sushshring: Change to `root["data"][0].asBool()` instead.

sushshring: Is this an array that will be returned? I assumed it would just be true or false? If it is a string, you'll have to compare it with something... I don't think it can just be cast directly to a bool.

httpwrapper/httpwrapper.h
```diff
@@ -0,0 +1,29 @@
...
+    String counter;
+    String numScans;
+  } jawn;
+
+  class HttpImpl{
+    private:
+      String redisHost;
+    public:
+      constructor(String host): redisHost(host);
+      jawn* getDataFromPin(int pin);
```
sushshring: Could we add docstrings on the methods?

sketch/hackPSUhttp.cpp
```diff
@@ -2,76 +2,73 @@
 #include "hackPSUhttp.h"
 
 namespace hackPSU {
-	static Response* HTTP::GET(String url, int count, Headers headers[]){
-		HTTPClient http;
-   		http.begin(url); //http Begin call
+  static Response* HTTP::GET(String url, int count, Headers headers[]){
```
sushshring: Thanks for fixing the spacing!

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,94 @@
...
+		
+
+		Response* response = HTTP.POST(url, payload, count, headers);
+
+		if (response->responseCode < 0){
+			throw "http request failed :(";
+		}
+
+		StaticJsonBuffer<200> jsonBuffer;
+		JsonObject& root = jsonBuffer.parseObject(response->payload);
```
sushshring: Response should be freed here once all its data has been parsed. Syntax: `delete response;`

httpwrapper/httpwrapper.h
```diff
@@ -0,0 +1,46 @@
...
+namespace hackPSU {
+  typedef struct redisData{
+    String uid;
+    String pin;
+    String name;
+    String shirtSize;
+    String diet;
+    String counter;
+    String numScans;
+  } jawn;
```
sushshring: Change to redisData here as well

sushshring:
This is starting to take great shape! Added some more small comments.

httpwrapper/httpwrapper.h
```diff
@@ -0,0 +1,48 @@
...
+    String uid;
+    String pin;
+    String name;
+    String shirtSize;
+    String diet;
+    String counter;
+    String numScans;
+  } redisData;
+
+  class HttpImpl{
```
sushshring: Could you include a doctoring at the class level showing example usage?

sushshring:
One final comment, LGTM otherwise. Will approve after sketch test.

sketch/hackPSUhttp.cpp
```diff
@@ -17,6 +17,11 @@ namespace hackPSU {
     responseInfo->payload = http.getString(); 		//getting the repsonse from the get
     responseInfo->responseCode = httpCode;
 
+    const char* errrorString []= {"success","HTTPC_ERROR_CONNECTION_REFUSED","HTTPC_ERROR_SEND_HEADER_FAILED","HTTPC_ERROR_SEND_PAYLOAD_FAILED", "HTTPC_ERROR_NOT_CONNECTED"};
+    int code = (httpCode*-1);
```
sushshring: cool 😎

sketch/hackPSUhttp.cpp
```diff
@@ -17,6 +17,11 @@ namespace hackPSU {
...
+    String x = errrorString[code];
+    responseInfo->errorStrings = x;
```
sushshring: Nitpick: `responseInfo->errorString`, since there is only one.

sushshring: Also could you assign to responseInfo->errorString directly?

sketch/httpwrapper.cpp
```diff
@@ -17,6 +17,7 @@ namespace hackPSU {
 			//throw "http request failed :(";
      Serial.println("get rekt :(");
```
sushshring: Same error handling here?

sushshring:
Since the same error handling code is repeated over and over, you could make it its own private function that you just call each time you deal with the responseCode. Something like `string handleResponseCode(int responseCode){...}`

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,106 @@
...
+		String url = "https://"+this.redisHost+"/tabs/getpin";
+		String payload = "{\"pin\":"+pin+"}";
+		int count = 1;	
+		Headers headers [] = { { "Content-Type", "application/json" } };
+		
+
+		Response* response = HTTP.POST(url, payload, count, headers);
+
+		if (response->responseCode < 0){
+			throw "http request failed :(";
```
sushshring: Since we're handling the error codes internally to the http request code, change this to print to Serial and return null instead of just throwing.

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,106 @@
...
+		pinData->counter = root["counter"].asString();
+		pinData->numScans = root["numScans"].asString();
+
+  		return pinData;
+	}
+
+	bool HTTPImpl::assignRfidToUser(String userId, String userBandId){ 
+
+		String url = "https://"+this.redisHost+"/tabs/setup";
+		String payload = "{\"userID\":"+userId+",\"userBandId\":"+userBandId+"}";
```
ThePBJain: for setup the payload should pass: 
{
    id: rfid code,
    pin: base 10 pin
}

httpwrapper/httpwrapper.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+  		//Redis json parse
+  		return root["status"].asString() == "success";
+	}
+
+	bool HTTPImpl::entryScan(String userBandId, String locationId){
+
+
+		String url = "https://"+this.redisHost+"/tabs/add";
+		String payload = "{\"userBandID\":"+userBandId+",\"locationId\":"+locationId+"}";
```
ThePBJain: the add endpoint requires this request:
{
    location: locationId,
    id: rfid tag
}

httpwrapper/sketch/sketch.ino
```diff
@@ -0,0 +1,47 @@
...
+#define USE_SERIAL Serial
+
+ESP8266WiFiMulti WiFiMulti;
+
+void setup(){
+  
+  USE_SERIAL.begin(115200);
+  // USE_SERIAL.setDebugOutput(true);
+
+  USE_SERIAL.println();
```
ThePBJain: Could probably consolidate these 3 lines into USE_SERIAL.println("\n\n");

httpwrapper/sketch/sketch.ino
```diff
@@ -0,0 +1,47 @@
...
+  USE_SERIAL.println();
+
+  for (uint8_t t = 4; t > 0; t--) {
+    USE_SERIAL.printf("[SETUP] WAIT %d...\n", t);
+    USE_SERIAL.flush();
+    delay(1000);
+  }
+
+   WiFi.begin("Sushrut's iPhone", "qazwsx123");
```
ThePBJain: lol, plz no keys in code.

sketch/httpwrapper.cpp
```diff
@@ -0,0 +1,110 @@
...
+		pinData->counter = root["counter"];
+		pinData->numScans = root["numScans"];
+
+  		return pinData;
+	}
+
+	bool HTTPImpl::assignRfidToUser(String userId, String userBandId){ 
+
+		String url = "https://"+redisHost+"/tabs/setup";
+		String payload = "{\"userID\":"+userId+",\"userBandId\":"+userBandId+"}";
```
ThePBJain: double check payload. Redis should receive:
{
    id: rfid code,
    pin: base 10 pin
}

sketch/httpwrapper.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+  		//Redis json parse
+  		return root["status"] == "success";
+	}
+
+	bool HTTPImpl::entryScan(String userBandId, String locationId){
+
+
+		String url = "https://"+redisHost+"/tabs/add";
+		String payload = "{\"userBandID\":"+userBandId+",\"locationId\":"+locationId+"}";
```
ThePBJain: double check payload. Redis should receive:
{
id: rfid code,
location: locationId
}

sketch/wifiCheck/wifiCheck.ino
```diff
@@ -0,0 +1,16 @@
+#include <ESP8266WiFi.h>
+
+const char* ssid = "julia's iPhone";
+const char* password = "p@ssw0rd";
```
ThePBJain: plz no keys!!

ThePBJain: could you see if you could make a environment variable file (".env") or something for these passwords?

sushshring: I don't think that's how it works for C++, but a constants file could be used to connect to the wifi. I'd recommend using serial input, but we don't have a full ascii keyboard. Plus this is a sketch file, so it won't end up on the actual library.

ThePBJain: What's our plan then to connect to the wifi? Also we don't have to worry about enterprise wifi right?

sushshring: I’ll let @mdh5389 answer the first one. We should have a regular WPA2 network for the scanners

sushshring:
This looks good to me!

PR #15, opened 2018-09-26
State machine

mcderdav:
Hello HackPSU RFID team,

This review is coming in a little early; we still have one known bug in the code, which should be resolved soon, but as this is a large PR, I wanted to get it started early.  Please actually review this code, don't just hit `send it`; we want to have good code.  The current state of the `state_machine` branch is tested and known to be working as of this first comment.  Please test all changes fully (not just compilation) before committing.  Please be verbose in your review and give clear suggestions and reasons for changes; many changes will likely be necessary.

Happy Reviewing,
Dave McDermott


sketch/hackPSUconfig.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef WIFI_CONSTS
+#error Wifi constants must be updated
+#endif
+
+namespace hackPSU{
+  constexpr char* REDIS  = "c-98-235-90-5.hsd1.pa.comcast.net:3000";
+  constexpr char* SSID   = "";
+  constexpr char* PASSWD = "";
+  // SHA1 fingerprint of the certificate
+  constexpr char* fingerprint = "AD:0E:A5:F9:AB:6A:EF:B1:25:3A:A4:47:3D:A5:75:1A:E9:8C:A7:B5";
```
ThePBJain: Can get rid of this.

sketch/sketch.ino
```diff
@@ -5,11 +5,8 @@ using namespace hackPSU;
 
 Box* box;
 
-#define REDIS  "127.0.0.1"
-#define SSID   "psuwirelesssetup"
-#define PASSWD ""
-
 void setup() {  
+  Serial.begin(9600);
   box = new Box(REDIS, SSID, PASSWD, DEV, nullptr);
```
ThePBJain: This will have to be set to PROD right?

sketch/writeUID/writeUID.ino
```diff
@@ -0,0 +1,27 @@
+#include "hackPSUrfid.h"
```
ThePBJain: Are these test files? Should we have a separate folder for tests?

sketch/writeUID/writeUID.ino
```diff
@@ -0,0 +1,27 @@
...
+  // put your setup code here, to run once:
+  scanner = new Scanner(D8, D4);
+  Serial.begin(115200);
+  Serial.println("HI THERE");
+}
+
+void loop() {
+  // put your main code here, to run repeatedly:
+  
+  Serial.println("SCAN ME UP BOI");
```
ThePBJain: lol why?

sketch/writeUID/hackPSUrfid.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <stdint.h>
+#include <MFRC522.h>
+#include <SPI.h>
+//#include <MFRC522Hack.h>
+
+namespace hackPSU {
+
+  class Scanner{
```
ThePBJain: Are all files in sketch/writeUID for a test or do they integrate back?

sketch/readWriteData/hackPSUconfig.h
```diff
@@ -0,0 +1,17 @@
+#ifndef CONFIG__H
+#define CONFIG__H
+
+namespace hackPSU{
+  constexpr char* REDIS  = "192.168.0.11:3000";//"c-98-235-90-5.hsd1.pa.comcast.net";
+  constexpr char* SSID   = "TheBlueWhat";
+  constexpr char* PASSWD = "The_Gr3@t-68";
```
ThePBJain: Remove all the passwords from this file.

sketch/httpwrapper.h
```diff
@@ -19,30 +20,36 @@ namespace hackPSU {
...
       /*
         Use case: The first time the bands are scanned during regisration and we need to assign the band to the user
         Parameters: String userId, String userBandId
         Returns: success or fail
      */
       bool assignRfidToUser(String rfidCode, String pin);
-      
+
       /*
         Use case: When the user is trying to scan into a workshop/event
```
ThePBJain: Mention food as well

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,196 @@
...
+        //data.get<char*>("name"); use if interested in displaying it down the road
+        if (root.get<String>("status") != "success"){
+          Serial.println("Failed: " + message);
+          return false;
+        }
 
-		//Free up memory since parsing is complete
-		delete response;
+        Serial.println("DECODED IT");
+        Serial.println(millis());
```
ThePBJain: What is this for?

sketch/hackPSUkeypad.h
```diff
@@ -3,13 +3,17 @@
 #include "hackPSUdisplay.h"
 
 #include <Arduino.h>
+#include<ESP8266WiFi.h>
+
+#include "hackPSUconfig.h"
```
ThePBJain: Do we need the config file for this class?

sketch/hackPSUdisplay.cpp
```diff
@@ -1,71 +1,107 @@
...
+        clear(row);
+        lcd->setCursor(0, row);
+        lcd->print(data[row].substring(data[row].length()-16));
+      } else {
+        lcd->print(msg);
+      }
     }
-    if(state == DEV || state == HEADLESS) {
-      Serial.println(msg);
+    if(mode == DEV || mode == HEADLESS) {
```
ThePBJain: What is headless mode?

sketch/hackPSUconfig.h
mcderdav: We need them; I removed my personal WiFi; we need to set it to the SSID/PASSWD from the event

sketch/sketch.ino
mcderdav: Any time we have calls to `Serial` methods we need this or we crash

sketch/writeUID/writeUID.ino
mcderdav: The subdirectories are examples/test files. The libraries in them need to be symlinked and moved up, yes

sketch/writeUID/writeUID.ino
mcderdav: Removing

sketch/hackPSUdisplay.cpp
mcderdav: Headless mode only prints to `Serial`; it is a debug mode for creating boxes to check if issues are with MCU or LCD

sketch/hackPSUkeypad.h
mcderdav: It doesn't hurt, but isn't used; removing for clean imports, good catch

sketch/httpwrapper.cpp
mcderdav: That was profiling the RTT; we proved what RTT is for this call, so we can remove them.  We should be aiming to remove unnecessary prints, so good catch.

sketch/hackPSUconfig.h
ThePBJain: no get rid of fingerprint. we use fp now.

sketch/sketch.ino
ThePBJain: Don't you have Mode_e for difference between DEV and PROD?

sketch/hackPSUdisplay.cpp
```diff
@@ -1,71 +1,107 @@
...
-        for(uint8_t i : I2C_ADDRESSES){
-          Wire.beginTransmission(i);
-          if(Wire.endTransmission() == 0){
-            lcd = new LiquidCrystal_I2C(i, 16, 2);
-      //      Serial.print("I2C LCD found at 0x" + i < 16 ? "0":"");
-      //      Serial.println(i, HEX);
-            break;
-          }
+  Display::Display(Mode_e mode){
+    this->mode = mode;
```
sushshring: Would an initializer list be better for code quality?

sketch/httpwrapper.cpp
mcderdav: FYI, decode time is 3ms, RTT is close to 600ms, so event network will clear timing issues

sketch/hackPSUdisplay.cpp
ThePBJain: cool.

sketch/hackPSUdisplay.cpp
```diff
@@ -1,71 +1,107 @@
...
+    data[0] = "";
+    data[1] = "";
+    
+    if(mode == PROD || mode == DEV){
+      const uint8_t I2C_ADDRESSES[] = {0x27, 0x3f};
+      Wire.begin(/*SDA, SCL*/);
+      for(uint8_t i : I2C_ADDRESSES){
+        Wire.beginTransmission(i);
+        if(Wire.endTransmission() == 0){
+          lcd = new LiquidCrystal_I2C(i, 16, 2);
```
sushshring: `this->lcd`? for consistency

sketch/httpwrapper.h
mcderdav: Added food and more detailed description of return value

sketch/hackPSUconfig.h
mcderdav: Ah got it, it shows SSID and PASSWD as well, removing the `char*` version

sketch/sketch.ino
mcderdav: Yes; I can push that into display, good call!

sketch/hackPSUhttp.cpp
```diff
@@ -4,16 +4,40 @@
...
+  // There is an httpclient function that helps HTTPClient::begin(String host, uint16_t port, String uri, bool https, String httpsFingerprint)
+  
+  Response* HTTP::GET(String url, int headerCount, Headers headers[]){
     HTTPClient http;
-    http.begin(url); //http Begin call
-
-    for (int i = 0; i<count; i++){
+    if (url.startsWith("https:")) {
+      http.begin(url, fp);
+    }else{
```
sushshring: nit: space before `else`

sketch/hackPSUkeypad.cpp
```diff
@@ -80,17 +81,20 @@ namespace hackPSU{
     for(int i = 0; i < maxLen && start + timeout > millis(); i++){
```
sushshring: Please describe this functionality in a docstring since there is very specific stuff going on.

sketch/hackPSUrfid.cpp
```diff
@@ -34,21 +46,114 @@ namespace hackPSU {
...
   }
 
-  void Scanner::setUID(uint32_t uid){
-    
-    //Create one hacky_boi to run the uid change
-    MFRC522Hack hacky_boi(&reader);
-    
-    //While we haven't set the UID, yield
-    while(!hacky_boi.MIFARE_SetUid((byte*)&uid, (byte)4, true))
+  bool Scanner::getData(byte* buffer, byte size, byte blockAddr, unsigned long timeout){
```
sushshring: This is a massive function. Should be broken apart into subroutines.

sketch/hackPSUrfid.cpp
```diff
@@ -34,21 +46,114 @@ namespace hackPSU {
...
+      yield();
+      if ((millis() - timeout > time) && timeout) {
+        //timeout
+        this->reader.PICC_HaltA();
+        return false;
+      }
+    }
+
+    //Wait for card with data and yield to coroutines while waiting
+    while(!this->reader.PICC_ReadCardSerial()){
```
sushshring: This is the same code as the previous block, could be merged into one function.

sketch/hackPSUrfid.cpp
```diff
@@ -34,21 +46,114 @@ namespace hackPSU {
...
+    }
+
+    cleanup_read:;
+    // Halt PICC
+    reader.PICC_HaltA();
+    // Stop encryption on PCD
+    reader.PCD_StopCrypto1();
+  }
+
+  bool Scanner::setData(byte* buffer, byte size, byte blockAddr, unsigned long timeout){
```
sushshring: Same

sketch/hackPSUstatemachine.cpp
```diff
@@ -1,62 +1,452 @@
...
+        return;        
+      }
+      char lid_buffer[10] = {0};
+      char uid_buffer[10] = {0};
+      Serial.println(uid);
+      if (uid != last_scan) {
+        if (http->entryScan(itoa(lid, lid_buffer, 10), itoa(uid, uid_buffer, 10))) {
+          display->print("Allow", 1);
+        } else {
+          //TODO LCD->print Deny
```
sushshring: Seems like this one is done?

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,196 @@
...
-			return false;
-		}
-
-
-
-		StaticJsonBuffer<200> jsonBuffer;
-		JsonObject& root = jsonBuffer.parseObject(response->payload);
+    bool HTTPImpl::getAPIKey(){
+
+        //TODO MAKE DAT ONE SECURE BOI
```
sushshring: Done?

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,196 @@
...
+        DynamicJsonBuffer jsonBuffer(response->payload.length());
+        JsonObject& root = jsonBuffer.parseObject(response->payload);
+
+        //Free up memory since parsing is complete
+        delete response;
+
+        //Redis json parse
+        String status = root["status"];
+        Serial.println(status);
+        String message = root["message"];  //Should message also be returned to display why user was not allowed in?
```
sushshring: Remove comment question

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,196 @@
...
+        JsonObject& data = root.get<JsonObject>("data");
+        apiKey = data.get<String>("apikey");
+        Serial.println(apiKey);
+        //The following is based on assumptions and should be checked
+        return (status == "success");
```
sushshring: This is an invalid assumption. Return based on actual error state.

sushshring: This should check the response code and see if it is a non 200/300 code.

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,196 @@
...
+        DynamicJsonBuffer jsonBuffer(response->payload.length());
+        JsonObject& root = jsonBuffer.parseObject(response->payload);
+
+        //Free up memory since parsing is complete
+        delete response;
+
+
+        if (String(root.get<char*>("status")) == "error")
+          return nullptr;
+//    if (!root.success()) {
```
sushshring: cleanup

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,196 @@
...
+            return false;
+        }
+
+        DynamicJsonBuffer jsonBuffer(response->payload.length());
+        JsonObject& root = jsonBuffer.parseObject(response->payload);
+
+        //Free up memory since parsing is complete
+        delete response;
+
+//    if (!root.success()) {
```
sushshring: cleanup

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,196 @@
...
+            //Free up memory since parsing is complete
+            delete response;
+            return false;
+        }
+
+        DynamicJsonBuffer jsonBuffer(response->payload.length());
+        JsonObject& root = jsonBuffer.parseObject(response->payload);
+
+        //Free up memory since parsing is complete
+        delete response;
```
sushshring: set response to nullptr everywhere. Bad memory is ur worst nightmare

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,196 @@
...
+
+//    if (!root.success()) {
+//      throw "json parsing failed :(";
+//      }
+
+        //Redis json parse
+        return root["status"] == "success";
```
sushshring: This is a poor representation of the actual response.

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,196 @@
...
+        Headers headers [] = { { "Content-Type", "application/json" } };
+
+
+        Response* response = HTTP::POST(url, payload, headerCount, headers);
+
+        if (response->responseCode < 0){
+            Serial.print("Http request failed: ");
+            Serial.println(HTTP::handleError(response->responseCode));
+            delete response;
+            return false;
```
sushshring: return `nullptr` not false.

sushshring:
This is looking a whole ton better than last week!

sketch/hackPSUdisplay.cpp
mcderdav: Can you do that when the member and the parameter are named the same?

sketch/hackPSUdisplay.cpp
mcderdav: In display convention is for no `this` references except `mode` because `mode` is a member and a constructor argument; Ideally, we remove `this` from mode, then no references to the `this` pointer exist in the class implementation

sketch/hackPSUdisplay.cpp
```diff
@@ -1,71 +1,107 @@
...
 
   void Display::print(char msg){
-    if(state == PROD || state == DEV){
-      lcd->print(msg);
+    data[row] += msg;
+    if(mode == PROD || mode == DEV){
+      if(data[row].length() > 16){
+        clear(row);
+        lcd->setCursor(0, row);
+        lcd->print(data[row].substring(data[row].length()-16));
```
jmm8046: So I'm assuming we're not implementing any type of scroll feature and that we're just losing the rest of the string if it is more than 16 characters?

sketch/hackPSUhttp.cpp
mcderdav: Addressed in nit commit, agree on spacing

sketch/hackPSUkeypad.cpp
mcderdav: @mdh5389 Is the original author, requesting changes from him directly

sketch/hackPSUdisplay.cpp
sushshring: In that case, initialize mode in the initializer list.

sketch/hackPSUdisplay.cpp
sushshring: yep you can

sketch/hackPSUrfid.cpp
mcderdav: Agreed; most of `getUID` is reproduced 3x; moving to C-style static implementation to share code.

sketch/hackPSUdisplay.cpp
sushshring: `mode(mode)`

sketch/hackPSUdisplay.cpp
jmm8046: I'm not too sure about this but I know it was mentioned to me that we need to free up memory at some point after creating objects, do we have to delete the lcd object here at some point or is it ok to just leave it?

sketch/httpwrapper.cpp
jmm8046: I have a commit on my branch in which this comment is removed

sketch/httpwrapper.cpp
jmm8046: Once again, this is changed on my branch

sketch/sketch.ino
ThePBJain: lit

sketch/httpwrapper.cpp
jmm8046: I'm gonna stop saying this now, but once again this is on my branch lol

sketch/httpwrapper.cpp
ThePBJain: true

sketch/hackPSUdisplay.cpp
ThePBJain: maybe later? @mcderdav

sketch/httpwrapper.cpp
ThePBJain: yea

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,182 @@
...
+        if (response->responseCode < 0){
+            Serial.print("Http request failed: ");
+            Serial.println(HTTP::handleError(response->responseCode));
+            Serial.println(response->errorMessage);
+            //Free up memory since parsing is complete
+            delete response;
+            return false;
+        }
 
+        DynamicJsonBuffer jsonBuffer(response->payload.length());
```
ThePBJain: Dynamic buffers are in heap. Shouldn't we be removing it from memory or does that automatically happen?

ThePBJain: https://arduinojson.org/v5/faq/how-to-reduce-memory-usage/ had some good points on how to reduce memory usage.

sketch/httpwrapper.cpp
ThePBJain: we might have to merge julias to ours? idk what's been going on in that branch

sketch/httpwrapper.cpp
ThePBJain: @sushshring

sketch/httpwrapper.cpp
mcderdav: Addressing now

sketch/httpwrapper.cpp
mcderdav: How should we address

sketch/httpwrapper.cpp
sushshring: Yeah PR that one into this if it is done

sketch/httpwrapper.cpp
sushshring: @ThePBJain

sketch/httpwrapper.cpp
mcderdav: Addressed on this branch; hopefully merge conflict won't be too bad with yours @jmm8046

sketch/hackPSUdisplay.cpp
mcderdav: For Release 1.0, we're not targeting a scroll feature; go in and make a ticket for that either in Asana or on GitHub @jmm8046.  I would love to see that by this hackathon @ThePBJain, but I don't know if it'll happen

sketch/httpwrapper.cpp
ThePBJain: Id be careful with that. it could break a lot of things

sketch/hackPSUdisplay.cpp
mcderdav: So in this case, it's okay to leave them because they persist for the lifetime of the program; technically, they should be deleted in the destructor for the class; I'll add one now

sketch/httpwrapper.cpp
mcderdav: I'm going to say for now, leave it because it works even though it uses a lot of memory.  We should add a task either on GitHub or Asana to follow that guide for next time

sketch/hackPSUdisplay.cpp
mcderdav: Lit, making that change now

sketch/hackPSUdisplay.cpp
mcderdav: Sent it!

mcderdav: Changes made near hackPSUdisplay.h:21 and hackPSUdisplay.cpp:33

sketch/hackPSUdisplay.cpp
mdh5389: @wutnut will be implementing this hopefully tonight

mcderdav: Beautiful, is he in this branch, or are we going to have another merge hell

sketch/sketch.ino
mcderdav: Idk why it hasn't updated here, but peep `sketch/sketch.ino` for the update

sketch/hackPSUstatemachine.cpp
```diff
@@ -1,62 +1,452 @@
...
+      break;
+    default:
+      state = LOCK;
+      break;
   }
+}
 
-  void Box::scan(void){
+/*
+
```
mdh5389: There should be comments here or the block comment should be removed

sketch/hackPSUstatemachine.cpp
```diff
@@ -1,62 +1,452 @@
...
+    case 4:
+      display->print("5:Lock", 1);
+      break;
+    default:
+      menu_state = 0;
+      state = LOCK;
+      return;
+  }
+
+  switch (keypad->getUniqueKey(500)) {
```
mdh5389: Is there any reason this timeout can't be increased?  There is nothing else which needs to be happening while the operator is taking their time to make a selection

sketch/hackPSUstatemachine.cpp
```diff
@@ -1,62 +1,452 @@
...
+      menu_state = (MENU_STATES + menu_state - 1) % MENU_STATES;
+      menu_state %= MENU_STATES;
+      break;
+    case 'B':
+      menu_state++;
+      menu_state %= MENU_STATES;
+      break;
+    case '1':
+      state = LOCATION;
+      menu_state = 0;
```
mdh5389: Should the be a method which "tears down" each state? So like menu_teardown in this case would assign menu_state = 0 and we could add anything else if there needs to be something added (like a display->clear() or something)

sketch/hackPSUstatemachine.cpp
```diff
@@ -1,62 +1,452 @@
...
+  // Do not select location based on a number
+  switch (keypad->getUniqueKey(500)) {
+    case 'A':
+      location_state++;
+      location_state %= num_locations;
+      break;
+    case 'B':
+      location_state = (num_locations + location_state - 1) % num_locations;
+      break;
+    case 'C':
```
mdh5389: For now, this should be before default and not have a break since it is not implemented (or fully commented out)

sketch/hackPSUstatemachine.cpp
mdh5389: Correct, comment should be nuked.

sketch/hackPSUstatemachine.cpp
```diff
@@ -1,62 +1,452 @@
...
+  display->print("Scan wristband", 0);
+  do {
+    display->clear(1);
+    
+    uid = scanner->getUID(SCAN_TIMEOUT);
+    if(!uid){
+      display->print("Scanner timeout", 1);
+      delay(1000);
+      display->clear(1);
+      return;
```
mdh5389: Why does this return?  I don't want to return and restart the process if the scanner times out, There should be keypad handling while no scan is read.  The keypad is the way out here,.

sketch/hackPSUstatemachine.cpp
```diff
@@ -1,62 +1,452 @@
...
+      display->print("Scrap wristband!", 1);
+      uid = 0;
+      delay(2000);
+    }
+    
+  } while(!uid);
+
+
+  display->print("Shirt Size: ", 0);
+  //TODO fix dat boi
```
mdh5389: Was this ever resolved @mcderdav ?

sketch/hackPSUstatemachine.cpp
```diff
@@ -1,62 +1,452 @@
...
+  display->print(data->shirtSize);
+
+  display->print("Photo consent?",1);
+
+  // Require keypress to continue
+  while(keypad->getUniqueKey(1000) != 't');
+
```
mdh5389: Does a display->clear() need to happen here?

sketch/hackPSUstatemachine.cpp
```diff
@@ -1,62 +1,452 @@
...
+  delete data;
+}
+
+void Box::wifi() {
   
+  display->print("B: RETURN", 0);
```
mdh5389: I put this here to show how to exit at the cost of not displaying the SSID.  Should we show the SSID or command controls?

sketch/hackPSUstatemachine.cpp
```diff
@@ -1,62 +1,452 @@
...
+
+  switch(keypad->getUniqueKey(100)){
+    case 'B': 
+      state = MENU;
+      break;
+    case 'D':
+      state = LOCK;
+      break;
+    default:
+      // TODO: write master wristband flow (if present)
```
mdh5389: All the code below should be moved here OR the previous case statements should return rather than break.  I would prefer the former.

sketch/hackPSUstatemachine.h
```diff
@@ -1,15 +1,20 @@
...
+#include<ESP8266WiFi.h>
 
 #include "httpwrapper.h"
 #include "hackPSUkeypad.h"
 #include "hackPSUdisplay.h"
 #include "hackPSUrfid.h"
+#include "hackPSUconfig.h"
+
 
 #define GOLDEN_KEY 0xC0DEBABE
```
mdh5389: Do we want to make the golden key something we can type into the keypad to confirm for security purposes?

sketch/httpwrapper.cpp
mdh5389: Do we want to remove DynamicBuffer or follow the guidance in the link? Commenting because the TODO earlier

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,182 @@
...
+            //Free up memory since parsing is complete
+            delete response; response = nullptr;
+            return nullptr;
+        }
 
+        DynamicJsonBuffer jsonBuffer(response->payload.length());
+        JsonObject& root = jsonBuffer.parseObject(response->payload);
+
+        //Free up memory since parsing is complete
+        delete response; response = nullptr;
```
mdh5389: Can this be on two lines please?

mdh5389: Would it be worth #define delete(a) delete a; a = nullptr; so we always make sure to adjust the pointer after it's deleted?

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,182 @@
...
+
+        //Redis json parse
+        String status = root["status"];
+        Serial.println(status);
```
mdh5389: I'd prefer not to have Serial debugging in prod.  It might be worth adding a debug function to the display class and we can set a output level (info, warn, error, debug, etc.)

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,182 @@
...
+        apiKey = data.get<String>("apikey");
+        Serial.println(apiKey);
+        //The following is based on assumptions and should be checked
+        return (status == "success");
+    }
+
+    redisData* HTTPImpl::getDataFromPin(String pin){
+        String url = "https://"+redisHost+"/tabs/getpin";
+        Serial.println(url);
+        String payload = "{\"pin\":"+pin+", \"apikey\": \""+apiKey+"\"}";
```
mdh5389: Coming off from the memory optomizations from the DynamicBuffer stuff, should this follow similar guidelines?

sketch/httpwrapper.cpp
```diff
@@ -1,114 +1,182 @@
...
+        delete response; response = nullptr;
+
+        //Redis json parse
+        return root["status"] == "success";
+    }
+
+    bool HTTPImpl::entryScan(String locationId, String rfidTag){
+
+
+        String url = "https://"+redisHost+"/tabs/add";
```
mdh5389: When making the URL can we have the "https://"+redisHost be something defined in the class so we just have to add the route here (this would also make only one place to change http/https depending on the needs)

sketch/readWriteData/hackPSUconfig.h
```diff
@@ -0,0 +1,17 @@
+#ifndef CONFIG__H
+#define CONFIG__H
+
+namespace hackPSU{
+  constexpr char* REDIS  = "192.168.0.11:3000";//"c-98-235-90-5.hsd1.pa.comcast.net";
```
mdh5389: Should we keep the IP or the hostname here?  I guess the IP would save a few bytes of memory...

mdh5389: Follow up from previous comment, should this be "https://192.168.0.11:3000".  Also for stuff that is always changing like IP addresses (actual vs dev) could we use #define for this.  Master should have the actual server not a local IP (@ThePBJain )

mdh5389: Wait WTF, why is this directory/file here? We should either start making an examples directory or nuke all non-essential directories within sketch. (@mcderdav ?)

sketch/hackPSUconfig.h
```diff
@@ -0,0 +1,28 @@
...
+
+//Uncomment this line if the SSID & PASSWDS fields are correct
+#define WIFI_CONSTS
+#ifndef WIFI_CONSTS
+#error Wifi constants must be updated
+#endif
+
+namespace hackPSU{
+  constexpr char* REDIS  = "c-98-235-90-5.hsd1.pa.comcast.net:3000";
+  constexpr char* SSID   = "McDermott";
```
mdh5389: In master, let's not actually have any personal SSID/passwords.

sketch/sketch.ino
mdh5389: Why was this added? It is handled in the display class

mcderdav: @mdh5389 This window is out of date; view the current code in GitHub's file explorer; it's removed

sketch/httpwrapper.cpp
mcderdav: @mdh5389 For now DynamicBuffer because it works, we can Asana it for next time

sketch/hackPSUdisplay.cpp
```diff
@@ -1,71 +1,111 @@
 #include "hackPSUdisplay.h"
```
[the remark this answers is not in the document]
mcderdav: @wutnut can you handle that one; you're already in this class

sketch/hackPSUdisplay.cpp
```diff
@@ -1,71 +1,111 @@
...
+        clear(row);
+        lcd->setCursor(0, row);
+        lcd->print(data[row].substring(0,16));
+      } else {
+        lcd->print(msg);
+      }
     }
-    if(state == DEV || state == HEADLESS) {
+    if(mode == DEV || mode == HEADLESS) {
       Serial.println(msg);
```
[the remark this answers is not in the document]
mcderdav: The implementation does not know cursor position; unless we implement a cursor tracker ourselves, we cannot give the row that we are printing on

sketch/hackPSUhttp.cpp
```diff
@@ -4,16 +4,40 @@
...
-  Response* HTTP::GET(String url, int count, Headers headers[]){
+  //TODO: Refactor all of this so that you pass in host, port and uri to make more extensible
+  // There is an httpclient function that helps HTTPClient::begin(String host, uint16_t port, String uri, bool https, String httpsFingerprint)
+  
+  Response* HTTP::GET(String url, int headerCount, Headers headers[]){
     HTTPClient http;
-    http.begin(url); //http Begin call
-
-    for (int i = 0; i<count; i++){
+    if (url.startsWith("https:")) {
```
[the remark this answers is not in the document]
mcderdav: I would prefer to leave the conditional; we never know what we're going to do and for the one extra branch it's probably not that important, but we should ping @ThePBJain and @sushshring on this call

sketch/hackPSUhttp.cpp
```diff
@@ -24,32 +48,42 @@ namespace hackPSU {
...
+      http.begin(url, fp);
+    }else{
+      http.begin(url); //http Begin call
+    }
 
-    for (int i = 0; i<count; i++){
+    Serial.println("HEAD");
+    for (int i = 0; i<headerCount; i++){
       Headers header = headers[i];
+      Serial.println(header.headerKey + ":" + header.headerValue);
```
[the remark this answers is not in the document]
mcderdav: @mdh5389 Good call, I set a reminder to do that for tomorrow at noon; I will search the code and drop if not in display.  Call me out if it's not done by tomorrow night

sketch/hackPSUrfid.cpp
```diff
@@ -3,52 +3,107 @@
...
     this->reader.PCD_Init();
     this->reader.PCD_SetAntennaGain(reader.RxGain_max);
     this->key = key;
   }
 
   Scanner::Scanner(uint8_t ssPin, uint8_t rstPin) : Scanner(ssPin, rstPin, Scanner::DEFAULT_KEY) {}
-  
-  uint32_t Scanner::getUID(void){
+
+  uint32_t Scanner::getUID_noStop(unsigned long timeout){
```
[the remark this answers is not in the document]
mcderdav: @mdh5389 Look at the implementations for get & set data; they require a getUID beforehand, but without the `PICC_HaltA()` call afterwards.  This was done to remove bloat in the codebase; it has been tested and is working as intended.  We can revert commmit `7228d68b44068cf452f1a1242474e6efdbc6e39a` if you disagree with this change.

sketch/httpwrapper.cpp
mcderdav: These need to be removed; they serve no purpose past debug; a reminder is set for me to remove them

sketch/httpwrapper.cpp
mcderdav: We don't really have any better way; we could return a gigantic object on the stack, but I don't mind this, the only thing that I disagree with is the multiple allocations from `String.concat()` here, but again, we don't really have a better way.  What are your thoughts on imrovements @mdh5389

sketch/httpwrapper.cpp
mcderdav: Good call, implementing now

mcderdav: Changed constructor to implement, committing now

sketch/readWriteData/hackPSUconfig.h
mcderdav: @mdh5389 never https://in front of the const.  Also this is a test case; this is the readWriteData example code.  Don't worry about this one here; we can drop it if you want

sketch/hackPSUconfig.h
mcderdav: Goddamnit, I thought I removed those, agreed

mcderdav: Removed personal WiFi and set to not compile if SSID/PASSWD are unset

sketch/sketch.ino
mdh5389: Then we gucci

sketch/hackPSUdisplay.cpp
mdh5389: He should be in another branch, I'll verify when I meet with him

sketch/httpwrapper.cpp
mdh5389: Does this have any impact on the malloc failures you were experiencing?

sketch/hackPSUdisplay.cpp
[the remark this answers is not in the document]
mdh5389: This is for Serial, not LCD.  It does actually know the row.  Unless someone changes it through the LCD object

sketch/hackPSUrfid.cpp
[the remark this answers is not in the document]
mdh5389: Okay, I see why you need it, carry on

sketch/httpwrapper.cpp
mdh5389: I was thinking in terms of a more uniform looking code so we might change these to char* to match everywhere else we need char* rather than String objects

sketch/readWriteData/hackPSUconfig.h
mdh5389: I'd prefer tests not be in the the final code.  The can be in a /test directory or with /examples

ThePBJain: @mdh5389 it should not be https://

ThePBJain: that's done elsewhere

sketch/httpwrapper.cpp
ThePBJain: It probably does. We don't really need DynamicBuffer for the other ones, just getLocation. Do you mind if we change it?

sketch/hackPSUhttp.cpp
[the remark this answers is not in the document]
ThePBJain: I'd leave the conditional. They actually have a built in conditional http.begin(url) function that I will switch to down the road. Read the TODO in the file to for more info.

sketch/hackPSUstatemachine.cpp
```diff
@@ -1,62 +1,452 @@
...
+  delete display;
+  Serial.println("Deleted all Box members");
+}
+
+void Box::cycle(void) {
+  //switch on state; default to init
+  switch (state) {
+    case LOCK:
+      lock();
+      break;
```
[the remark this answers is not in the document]
sushshring: Preferred returns if possible.

sketch/httpwrapper.cpp
sushshring: I like that idea

sketch/hackPSUdisplay.cpp
mcderdav: Lit, we'll handle merge hell after this then

sketch/httpwrapper.cpp
mcderdav: @ThePBJain you're free to update and test; avoiding a malloc call will also reduce the time to process json responses (currently 3ms).  I personally don't care, but if you're passionate about it for this release, make the changes and do a full functionality test before pushing.

sketch/hackPSUdisplay.cpp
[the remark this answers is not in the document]
mcderdav: Totally missed that `Display` has a `row` member; updating now

sketch/hackPSUhttp.cpp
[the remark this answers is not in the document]
mcderdav: Sent it

sketch/httpwrapper.cpp
mcderdav: These have to be strings to allow the + for concat; otherwise it's trying to add the literal `char *`s and it starts to do pointer arithmetic

sketch/readWriteData/hackPSUconfig.h
mcderdav: @mdh5389 Good call; permission to nuke them for now?

sketch/httpwrapper.cpp
ThePBJain: assigns rfid if is successful. 
return root.get<String>("status") == "success";

sketch/hackPSUkeypad.cpp
mdh5389: @sushshring I have this type of comment in the .h file but I am adding some comments into the .cpp file itself for the less intuitive things.

sketch/hackPSUkeypad.cpp
```diff
@@ -80,17 +81,20 @@ namespace hackPSU{
...
+      }else if( key != 't' && key != 'z') {
+        display->print(key);
         pin += key;
         start = millis(); // timeout is from the last pressed key
       } else {
         i--;
+        yield();
       }
     }
     if( start + timeout < millis() ){
```
mdh5389: There should be some kind of indication that the pin exceeded expected length rather than returning the pin up to the max length

sketch/hackPSUkeypad.cpp
sushshring: 👍

sketch/httpwrapper.cpp
sushshring: 👍

sketch/httpwrapper.cpp
sushshring: @ThePBJain Isn't that the same thing? Didn't you say status is always success as long as the request arrived?

sketch/httpwrapper.cpp
sushshring: Do we have access to `asprintf()`? That takes care of safe string concatenation.

sketch/readWriteData/hackPSUconfig.h
sushshring: nuke it

sketch/httpwrapper.cpp
mcderdav: No `asprintf`, just `sprintf`.  Issue is that pin & api key are `String`, so they need to be downcast to `c_str` before they can be used there.  Using implicit `concat` via `+` operator is the correct `C++`ism.  If we want this cleaner, we can move to `payload = "{\"pin\":\"$PIN\", \"apikey\":\"$API_KEY\"}"` and use `String.replace()` on it

sketch/readWriteData/hackPSUconfig.h
mcderdav: 🚀 nukt

sketch/httpwrapper.cpp
ThePBJain: Yea which is fine for assigning RFID tags. Will return error if isn't assigned. But it won't let you know if it didn't go to api.

sushshring:
Duplicate state menu doesnt work.

mcderdav:
Reminder comment; we should fix redisData implementation to not have pointers to stack

PR #16, opened 2018-09-28
HTTP Corrections

sushshring:


sketch/httpwrapper.cpp
```diff
@@ -20,10 +20,15 @@ namespace hackPSU {
         Response* response = HTTP::POST(url, payload, headerCount, headers);
 
         if (response->responseCode < 0){
+            Serial.print("Http request failed: ");
```
sushshring: Print statements have been nuked. Get rid of these

sketch/httpwrapper.cpp
```diff
@@ -114,8 +130,12 @@ namespace hackPSU {
...
+            Serial.println(HTTP::handleError(response->responseCode));
+            Serial.println(response->errorMessage);
             //Free up memory since parsing is complete
-            delete response; response = nullptr;
+            delete response;
+            response = nullptr;
```
ThePBJain: Why return nullptr when expecting bool response?

sketch/httpwrapper.cpp
```diff
@@ -114,8 +130,12 @@ namespace hackPSU {
         Response* response = HTTP::POST(url, payload, headerCount, headers);
 
         if (response->responseCode < 0){
+            Serial.print("Http request failed: ");
+            Serial.println(HTTP::handleError(response->responseCode));
+            Serial.println(response->errorMessage);
```
ThePBJain: Did nothing with the data. Need to check if isRepeat is true or not.

sushshring:
In httpwrapper.cpp, change all instances of String which don't have concatenation to use char *

sketch/httpwrapper.cpp
```diff
@@ -47,23 +52,31 @@ namespace hackPSU {
...
         JsonObject& root = jsonBuffer.parseObject(response->payload);
 
         //Free up memory since parsing is complete
-        delete response; response = nullptr;
+        delete response; 
+        response = nullptr;
 
 
-        if (String(root.get<char*>("status")) == "error")
+        if (String(root.get<char*>("status")) == "error") {
```
ThePBJain: strncmp(root.get<char *>("status"), "success", 7)

mdh5389:
We will need to change hackPSUstatemachine.cpp to act on each of the responses for all the modified functions.  Also, can we get an update on the documentation on the headers (particularly describe what they return)?

sketch/httpwrapper.h
```diff
@@ -8,6 +8,8 @@
 #include "hackPSUconfig.h"
 
 namespace hackPSU {
+  typedef enum {SUCCESS, FAIL, TIMEOUT, REDIS_DOWN} api_response;
```
mcderdav: This will not compile; the RFID status enum also uses TIMEOUT; change one or the other so that there are no naming collisions

sketch/httpwrapper.h
```diff
@@ -35,7 +37,7 @@ namespace hackPSU {
       String redisHost, apiKey;
     public:
       
-      bool getAPIKey();
+      api_response getAPIKey();
```
mcderdav: With these new prototypes, we need to validate responses differently in the state machine file; it will compile as is, but without changes there, it will not work; our check is no longer `if(foo())`, it is now `if (foo() == SUCCESS)`

sketch/httpwrapper.h
mdh5389: Lets namespace all the enums to avoid conflicts

PR #21, opened 2018-10-04
Display functions

mdh5389:


sketch/hackPSUstatemachine.cpp
```diff
@@ -250,15 +250,18 @@ void Box::location_cleanup(){
     delete location_list;
     location_list = nullptr;
   }
+  num_locations = 0;
   location_state = 0;
   display->clear();
 }
 
 void Box::scan() {
+  display->print('*',NONE_C, '#', NONE_C, '\0', NONE_C, 'D', LOCK_C);
```
mcderdav: What does having null here do?  I don't see documentation on that anywhere

sketch/hackPSUstatemachine.cpp
```diff
@@ -337,17 +339,17 @@ void Box::checkin() {
     switch(keypress){
       case '*': // Wrong person
         if( !validated){
-          if (data != nullptr) delete data;
+          delete data;
```
mcderdav: Good catch; I forgot cpp `delete` handles null checks (unlike some `free` implementations)

sketch/hackPSUstatemachine.h
```diff
@@ -14,7 +14,7 @@
 
 
 #define GOLDEN_KEY 0xC0DEBABE
-//#define SECURE_BOX
+#define SECURE_BOX
```
mcderdav: Thank you for fix on that

sketch/hackPSUdisplay.cpp
```diff
@@ -62,32 +99,126 @@ namespace hackPSU {
...
+      lcd->home();
+      lcd->print("                ");
+      lcd->home();
+      for(const Menu &m: menu){
+        print(m);
+      }
+    }
+  }
+  
+  void Display::print(Menu control){
```
mcderdav: It looks like behavior of `\0` will be to print control and no leading char; is that what happens?

sketch/hackPSUdisplay.cpp
```diff
@@ -1,5 +1,11 @@
 #include "hackPSUdisplay.h"
 
+#if defined(ARDUINO) && ARDUINO >= 100
+#define printByte(args)  write(args);
```
mcderdav: Can we get at the end of this file a 
``` 
#ifdef printByte(args)
#undef printByte(args)
#endif 
```
That will prevent the macro from leaking out of this file

sketch/hackPSUdisplay.cpp
```diff
@@ -62,32 +99,126 @@ namespace hackPSU {
...
+            lcd->print(msg.substring(0,16));
+          } else {
+            lcd->print(msg);
+          }
       }
       if(mode == DEV || mode == HEADLESS) {
         Serial.println(msg);
       }
-      data[row] = String(msg);
+      data[row] = msg;
```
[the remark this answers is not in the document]
mdh5389: Not following you here.  Can you be more detailed?

sketch/hackPSUdisplay.h
```diff
@@ -4,17 +4,27 @@
 #include <Arduino.h>
 #include <LiquidCrystal_I2C.h>
 
-#include "hackPSUconfig.h"
+//#include "hackPSUconfig.h"
```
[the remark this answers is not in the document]
mdh5389: To test in a different sketch.  Will fix

sketch/hackPSUdisplay.h
```diff
@@ -4,17 +4,27 @@
...
 
 namespace hackPSU {
 
   typedef enum {PROD, DEV, HEADLESS} Mode_e;
+  typedef enum {CHECK_C, UP_C, DOWN_C, CLEAR_C, BACK_C, LOCK_C, SCROLL_C, NONE_C} Custom_char;
+  struct Menu{
```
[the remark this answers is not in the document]
mdh5389: These are the names of the characters we have.  the _C part is intended to be short for character.  We can change if need be

sketch/hackPSUstatemachine.cpp
```diff
@@ -155,13 +155,13 @@ void Box::menu() {
       menu_cleanup();
       break;
     case '5':
-      state = ZEROIZE;
-      menu_cleanup();    
-      break;
+      state =
```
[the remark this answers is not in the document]
mdh5389: No idea why that doesn't show ZEROIZE... Must be a GitHub thing

sketch/hackPSUstatemachine.cpp
mdh5389: So since the argument following is `NONE_C` it doesn't matter what character is there.  Will add documentation

sketch/hackPSUstatemachine.cpp
mdh5389: So at this point it is guaranteed to NOT be nullptr so regardless the check was pointless.

sketch/hackPSUdisplay.cpp
mdh5389: There's a check if the symbol is `NONE_C` it will print 4 spaces (clearing anything from that space)

sketch/hackPSUdisplay.cpp
mdh5389: Sure thing

sketch/hackPSUdisplay.cpp
```diff
@@ -18,13 +24,44 @@ namespace hackPSU {
...
+      lcd->createChar(SCROLL_C, scroll);
+      
       lcd->backlight();
-      lcd->setCursor(0,0);
+      lcd->home();//custom characters only work with lcd->home for some reason
     }
+
     if(mode == DEV || mode == HEADLESS) {
-      Serial.begin(BAUD_RATE);
+      Serial.begin(9600);
```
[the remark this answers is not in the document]
mdh5389: Shouldn't have.  Will re-add it.

sketch/hackPSUdisplay.cpp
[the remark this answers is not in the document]
mcderdav: You have a condition for `data[row] != msg` above; you only need to assign when that condition is met.  It doesn't really matter hence the nit prefix, but it saves a memory write

sketch/hackPSUdisplay.h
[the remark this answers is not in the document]
mcderdav: Thank you

sketch/hackPSUdisplay.h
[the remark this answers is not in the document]
mcderdav: Not the enum, but the struct; it represents a control not a menu.  I caught that the _C was Hungarian notation

sketch/hackPSUstatemachine.cpp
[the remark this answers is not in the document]
mcderdav: Okay, I'll buy that GitHub is borked

sketch/hackPSUstatemachine.cpp
mcderdav: Okay, thank you, that is what I thought, but I wasn't sure

sketch/hackPSUdisplay.cpp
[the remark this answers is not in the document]
mdh5389: It does only assign when that condition is met.  It is within that if statement

sketch/hackPSUdisplay.cpp
mcderdav: Got it, I dig that implementation, I must have missed the check, I'll go through again

sketch/hackPSUdisplay.cpp
mcderdav: Thank you!

sketch/hackPSUdisplay.cpp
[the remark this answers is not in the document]
mcderdav: Fak, I'm blind.  GG

PR #22, opened 2018-10-04
Julia/http corrections

mdh5389:


mdh5389:
I am very certain this code does not compile in its original state.  DO NOT MERGE until this is resolved

sketch/hackPSUstatemachine.cpp
```diff
@@ -364,7 +362,7 @@ void Box::checkin() {
       display->print("Scan wristband", 0);
       display->clear(1);
       uid = scanner->getUID(SCAN_TIMEOUT);
-      if( uid && !http->assignRfidToUser(String(uid), pin)) {
+      if( uid && http->assignRfidToUser(String(uid), pin) != responses::SUCCESS) {
```
sushshring: Will this start telling us to scrap wristbands if Redis goes down now?

mdh5389: Yes it will.  I still have testing to do I was just getting the ball rolling here.  I will put assignRfidToUser in a switch statement and go from there

PR #23, opened 2019-01-20
Matt/http combine

mdh5389:
Kept sketch/ for legacy sake.  All future development should be done in src/ and lib/ as appropriate.


ThePBJain:
What did you do to the code that was in httpwrapper? did it get moved to another file?

mdh5389:
It got collapsed into the network and requestclasses. I don't have some of the methods right now because I was waiting to test them. All the methods in the wrapper only differed in the HTTP action (get or post), the endpoint, or the specific data it fetched from the reaponse. So I tried combining all similar parts of the request.

PR #24, opened 2019-03-13
Thepbjain/network update

ThePBJain:
Adding and writing functions to connect with server. 
Please remember to check for coding style and naming schemes.
I also got rid of a couple files. let me know if I should add them back.


hackpsu.config
```diff
@@ -0,0 +1,7 @@
+; SSID of network to use
+  -DNETWORK_SSID=\"co.space\"
```
mdh5389: Please remove actual WiFi creds

lib/network/src/network.cpp
```diff
@@ -5,20 +5,22 @@
...
 
   //___________________________________________________________________ Request
-  Request::Request(API::Method method, String host, String route) : 
-      header(bf_header.createObject()), 
+  Request::Request(API::Method method, String host, String route) :
+      header(bf_header.createObject()),
       payload(bf_payload.createObject()),
       method(method)
   {
+    //Serial.println(host + route);
```
mdh5389: Please remove all commented out print statements or enclose then in:
`
#ifdef DEBUG
Serial.print()
#endif
`

lib/network/src/network.cpp
```diff
@@ -36,30 +38,36 @@ namespace hackPSU {
   }
 
   Response* Request::commit() {
+    String builtURL = url;
+    //Add payload to url if GET method
+    if(method == API::GET){
+      builtURL += "?";
```
mdh5389: So you are using the payload variable in GET requests as parameters?  This is okay as long as none of the other HTTP methods require a parameter to be set

lib/network/src/network.cpp
```diff
@@ -36,30 +38,36 @@ namespace hackPSU {
   }
 
   Response* Request::commit() {
+    String builtURL = url;
```
mdh5389: Why not keep appending to URL rather than making a new variable and sending the new variable?

lib/network/src/network.cpp
```diff
@@ -105,11 +113,17 @@ namespace hackPSU {
 
   Network::Network(String host): host(host) {
     req = nullptr;
-    WiFi.begin(SSID, PASSWD);
+    //apiKey = "805bc162-7d21-40cc-bc9b-1e0722c3ff88";
```
mdh5389: Encapsulate in 
`
#ifdef
 DEBUG 
#endif
`
or remove

lib/network/src/network.cpp
```diff
@@ -105,11 +113,17 @@ namespace hackPSU {
...
+    Serial.print("The host is: ");
```
mdh5389: Please remove or encapsulate these unnecessary print statements

lib/network/src/network.cpp
```diff
@@ -133,28 +147,150 @@ namespace hackPSU {
...
     if(registerScanner->code == 200){
       MAKE_BUFFER(25, 25) bf_data;
-      JsonObject& data = bf_data.parseObject(registerScanner->payload);
-
+      JsonObject& response = bf_data.parseObject(registerScanner->payload);
 
+      JsonObject& data = response.get<JsonObject>("data");
       apiKey = data.get<String>("apikey");
-
+      Serial.println(apiKey);
```
mdh5389: Remove/encapsulate

lib/network/src/network.cpp
```diff
@@ -133,28 +147,150 @@ namespace hackPSU {
...
     }
 
-    delete registerScanner;
     return res;
   }
+
+  String Network::getDataFromPin(int pin) {
+    Request* req = createRequest(API::POST, "/rfid/getpin");
+    req -> payload.set("pin", String(pin));
+    req -> payload.set("version", API_VERSION);
```
mdh5389: Does REDIS need this field?

lib/network/src/network.cpp
```diff
@@ -133,28 +147,150 @@ namespace hackPSU {
...
+    if(registerScanner->code == 200){
+      MAKE_BUFFER(25, 25) bf_data;
+      JsonObject& response = bf_data.parseObject(registerScanner->payload);
+      JsonObject& data = response.get<JsonObject>("data");
+      name = data.get<String>("name");
+      res = data.get<String>("version") == API_VERSION ? API::SUCCESS : API::OUTDATED;
+    } else {
+      res = static_cast<API::Response>(registerScanner->code);
+    }
+    Serial.print("Response Code: ");
```
mdh5389: Remove/encapsulate

lib/network/src/network.cpp
```diff
@@ -133,28 +147,150 @@ namespace hackPSU {
...
+    return name;
+  }
+
+  String Network::assignUserWID(int pin, String wid) {
+    Request* req = createRequest(API::POST, "/rfid/assign");
+    req -> payload.set("pin", String(pin));
+    req -> payload.set("wid", wid);
+    req -> payload.set("apikey", apiKey);
+    Response* registerScanner = req->commit();
+    String message = "NULL";
```
mdh5389: What is "NULL" the default value here?

lib/network/src/network.cpp
```diff
@@ -133,28 +147,150 @@ namespace hackPSU {
...
+      res = data.get<String>("version") == API_VERSION ? API::SUCCESS : API::OUTDATED;
+    } else {
+      res = static_cast<API::Response>(registerScanner->code);
+    }
+    Serial.print("Response Code: ");
+    Serial.println(res);
+    return name;
+  }
+
+  String Network::assignUserWID(int pin, String wid) {
```
mdh5389: Why does this function return a String not a bool indicating success?

lib/network/src/network.cpp
```diff
@@ -133,28 +147,150 @@ namespace hackPSU {
...
+    API::Response res;
+    if(registerScanner->code == 200){
+      MAKE_BUFFER(25, 25) bf_data;
+      JsonObject& response = bf_data.parseObject(registerScanner->payload);
+      message = response.get<String>("message");
+      res = response.get<String>("version") == API_VERSION ? API::SUCCESS : API::OUTDATED;
+    } else {
+      res = static_cast<API::Response>(registerScanner->code);
+    }
+    Serial.print("Response Code: ");
```
mdh5389: Remove/encapsulate

lib/network/src/network.cpp
```diff
@@ -133,28 +147,150 @@ namespace hackPSU {
...
+    Serial.println(res);
+    return message;
+  }
+
+  String Network::userInfoFromWID(String wid) {
```
mdh5389: Can we make a new srtuct for this method so no parsing later is needed? or is this only supposed to get the user's name?

lib/network/src/network.cpp
```diff
@@ -133,28 +147,150 @@ namespace hackPSU {
...
+      res = static_cast<API::Response>(registerScanner->code);
+    }
+
+    Serial.print("Response Code: ");
+    Serial.println(res);
+    Locations loc = {.data = nullptr, .length = 0};
+    return loc;
+  }
+
+  User Network::sendScan(String wid, int loc) {
```
mdh5389: Does every scan return the data in "User" and is this always needed? Can success be inferred by a bool/int and have a separate method to get this info?  Scanning into workshops doesn't need user data, just success/allow

lib/network/src/network.h
```diff
@@ -15,6 +15,11 @@
 
 namespace hackPSU {
 
+  /*typedef struct {
```
mdh5389: Can this be deleted if its not used?

platformio.ini
```diff
@@ -12,37 +12,37 @@
...
 
 [env:previous]
 platform = espressif8266
 board = nodemcuv2
 framework = arduino
-lib_deps = 
-  MFRC522@1.3.6
-  LiquidCrystal_I2C@1.1.4
+lib_deps =
+;  MFRC522@1.3.6
```
mdh5389: NO ACTION NEEDED
To use these, add Wire and SPI here

src/main.cpp
```diff
@@ -1,13 +1,62 @@
-#include "fullbox.h"
+#include <network.h>
```
mdh5389: Soon we should make controllers which will test specific things then we can choose which controller to run rather than always messing with main.cpp to test.

lib/network/src/network.cpp
ThePBJain: We only do get and post right

lib/network/src/network.cpp
ThePBJain: url is a global variable and we don't wanna mess with that.

lib/network/src/network.cpp
ThePBJain: question. would it look like this?
`#ifdef DEBUG
    Serial.print("The host is: ");
    Serial.println(host);
    Serial.println(NETWORK_SSID);
    Serial.println(NETWORK_PASSWORD);
    #endif`

lib/network/src/network.cpp
ThePBJain: what would you prefer?

lib/network/src/network.cpp
ThePBJain: Yea but the idea last time was to also include names so we can showcase that information; make it a little more personal. Either way it at least enables us to work with that information down the road. I remember seeing a todo comment in our old code about enabling that feature down the road.

src/main.cpp
ThePBJain: Yea. Wasn't sure how to set that up right now, but I'll leave that up to you.

lib/network/src/network.cpp
ThePBJain: We could, but the idea was that we could show the message that Redis returns if it fails. Redis could say things like: `Wristband Tag already opened.`, `Invalid pin`, `Redis database is down`. If we want to check for success, we only have to compare it to: `Created tab.`

ThePBJain: Let me know your thoughts on how you want to handle this @mdh5389

hackpsu.config
ThePBJain: should I just leave them blank

mdh5389: Or use hackpsu for both

PR #27, opened 2019-03-18
Revamp

mdh5389:


lib/network/src/network.cpp
```diff
@@ -1,160 +1,399 @@
...
-  bool Request::addPayload(String key, String value){
+  bool Network::Request::addPayload(String key, String value){
     return payload.set(key, value);
   }
-  bool Request::addHeader(String key, String value){
+  bool Network::Request::addHeader(String key, String value){
     return header.set(key, value);
   }
 
-  Response* Request::commit() {
```
ThePBJain: Might make sense to rename commit to send

lib/scanner/src/hackpsu/fullbox/fullbox.cpp
```diff
@@ -292,15 +292,13 @@ void Box::scan() {
...
-        JsonObject& scanData = bf_scan.createObject();
-        scanData.set("isRepeat", false);
-        scanData.set("status", "something");
-        /*if (http->entryScan(itoa(lid, lid_buffer, 10), itoa(uid, uid_buffer, 10), scanData)) {
+        User data = http->sendScan(String(uid), lid);
+        if (data.allow) {
           display->print("Allow", 1);
+          display->toggleDisplay();
         } else {
           display->print("Deny", 1);
```
ThePBJain: We need to explain why denied

lib/network/src/network.cpp
mdh5389: Would transmit be a better name?  
Also this class should strip out all the NodeMCU specific stuff (ESP8266WiFi.h and the commit method should become virtual)

lib/scanner/src/hackpsu/fullbox/fullbox.cpp
mdh5389: When should the toggle display happen?  In runtime, it seemed like I should do that after a successful reading of the RFID band, not after the HTTP request is returned.  This only makes a difference when the network is slow.  There were a bunch of people who were doing crazy things waiting for the scan to happen.

404 - unknown user
401 - get new api key
< 0 - network error, try again
>500 unknown error
Any others?

lib/network/src/network.cpp
mdh5389: I'm gonna push this to a different release.  There' gonna need to be a bunch of changes to this class to support specific board implementations for sending/receiving network data (Wi-Fi vs ethernet)

lib/scanner/src/hackpsu/fullbox/fullbox.cpp
ThePBJain: 409 is already assigned. 417 is expectation failed (basically used similarly to 404)

lib/network/src/network.cpp
ThePBJain: > Also this class should strip out all the NodeMCU specific stuff (ESP8266WiFi.h and the commit method should become virtual)

Make a virtual Send method.

lib/network/src/network.h
```diff
@@ -3,38 +3,68 @@
...
-    } Response;
-
     typedef enum {
       GET, POST
     } Method;
   }
 
-  typedef struct Response {
-    String payload;
+  class HTTPCode{
```
ThePBJain: Should we move classes like these outside of the network class, or is it fine to leave these here?

lib/network/src/network.h
```diff
@@ -3,38 +3,68 @@
...
 
-    operator String() const{
+    String toString() {
+      Serial.println("HTTP code: " + String(code));
       switch(code){
         // List of known/common http codes:
         case 200: return F("Success");
         case 401: return F("Unauthorized");
         case 404: return F("Not Found");
+        case 409: return F("Wristband Tag already opened");
```
ThePBJain: Use these to help you print out the correct statement

lib/network/src/network.h
```diff
@@ -50,81 +80,122 @@ namespace hackPSU {
...
+      bool addHeader(String key, String value);
+      bool addParameter(String key, String value);
+
+      bool checkApiKey();
+      HTTPCode getApiKey(int pin);
+      User getDataFromPin(int pin);
+      HTTPCode assignUserWID(String wid, int pin);
+      User userInfoFromWID(String wid);
+      Locations* getEvents();
+      User sendScan(String wid, String loc);
```
ThePBJain: I know we put this on here because we didn't have time, but we should move all these functions from line 143 to 151 to a subclass of network.

lib/network/src/network.h
```diff
@@ -50,81 +80,122 @@ namespace hackPSU {
...
+      
+      Items* getItems();
+      HTTPCode itemCheckout(String wid, int iid);
+      HTTPCode itemReturn(String wid, int iid);
```
ThePBJain: We don't do item return.

lib/network/src/network.cpp
```diff
@@ -1,160 +1,399 @@
...
+      payload.printTo(pld);
+      tmpcode = http.POST(pld);
+    }
+    if(tmpcode == 401){
+      Serial.println("Authentication error");
+      char apibuff[36] = {0};
+      EEPROM.put(0, apibuff);
+      EEPROM.commit();
+      if(reboot){
+        Serial.println("Rebooting...");
```
ThePBJain: shouldn't this be in a debug statement?

lib/network/src/network.cpp
mdh5389: Yes, but this will be a different release.  I am 100% aware of this.  If we keep adding changes to this PR it will never be merged

lib/network/src/network.cpp
```diff
@@ -1,160 +1,399 @@
...
+
+  void Network::createRequest(API::Method method, String route){
     if(req != nullptr) delete req;
+    req = nullptr;
     req = new Request(method, host, route);
-    return req;
   }
 
-  Response* Network::commitRequest(){
+  Response* Network::completeRequest(){
```
ThePBJain: why we need this?

lib/network/src/network.h
mdh5389: I'd say leave it for now, though it probably should be moved later

lib/network/src/network.cpp
ThePBJain: tis the bane of PRs.

lib/network/src/network.h
mdh5389: I think these should eventually make it into a different class which extends network

lib/network/src/network.h
mdh5389: Should it comment it out for now?  Eventually, we will be doing this

lib/network/src/network.cpp
mdh5389: Yeah... you got me there

lib/network/src/network.h
ThePBJain: yea

lib/network/src/network.h
ThePBJain: sure

lib/network/src/network.cpp
ThePBJain: lol

lib/network/src/network.cpp
ThePBJain: we never use this

mdh5389: Removed

lib/network/src/network.cpp
mdh5389: Fixed

PR #32, opened 2019-04-04
Net abstract

mdh5389:


bin/espota.py
```diff
@@ -0,0 +1,329 @@
+#!/usr/bin/env python
+#
+# Original espota.py by Ivan Grokhotkov:
+# https://gist.github.com/igrr/d35ab8446922179dc58c
+#
+# Modified since 2015-09-18 from Pascal Gollor (https://github.com/pgollor)
+# Modified since 2015-11-09 from Hristo Gochkov (https://github.com/me-no-dev)
+# Modified since 2016-01-03 from Matthew O'Gorman (https://githumb.com/mogorman)
+#
+# This script will push an OTA update to the ESP
```
ThePBJain: Didn't know that this could run python

mdh5389: So this is the script that pushes our an update for the nodemcu's.  It is here so if we ever decide to integrate updates with the redis server, we can actually do that.  This script doesn't get called from the scanners/platformio

ThePBJain: dope!

lib/network/src/api.cpp
```diff
@@ -0,0 +1,341 @@
...
+        apibuff[36] = '\0';
+        EEPROM.put(0, apibuff);
+        EEPROM.commit();
+    }
+
+    int Api::checkApiKey() {
+        if( !isAlphaNumeric(apiKey[0]) ) {
+            return 401;
+        }
+        Request* req = createRequest(API::POST,  "/auth/scanner/register");
```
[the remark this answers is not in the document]
mdh5389: should I also be changing all the /rfid to /scanner?

lib/network/src/api.cpp
```diff
@@ -0,0 +1,341 @@
...
+                    res->code = API_PARSE_ERROR;
+                }
+            } else {
+                res->code = API_PARSE_ERROR;
+            }
+        }
+        return cleanup(req, res);
+    }
+
+    int Api::getUserInfo(Fields field, String id, User* user){
```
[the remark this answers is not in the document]
mdh5389: So 'Fields' is an enum PIN or WID so this method can look up a user by either

lib/network/src/api.cpp
```diff
@@ -0,0 +1,341 @@
...
+        return cleanup(req, res);
+    }
+
+    int Api::getEvents(List<Event>* list) {
+        Request* req = createRequest(API::GET, "/rfid/events");
+
+        Response* res = sendRequest(req);
+
+        if(bool(*res)){
+            MAKE_BUFFER(25, 25) bf_data;
```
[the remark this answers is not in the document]
mdh5389: Currently I have it set up to use a dynamic buffer.  I don't know what an appropriate size for this buffer is.

lib/network/src/api.cpp
```diff
@@ -0,0 +1,341 @@
...
+        // req->addPayload("wid", wid);
+        // req->addPayload("itemId", String(item->id));
+
+        // Response* res = sendRequest(req);
+
+        // return cleanup(req, res);
+    }
+
+
+    void Api::pre_send(Request* request){
```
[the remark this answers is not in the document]
mdh5389: The network class has this as a virtual method pre_send and post_send that is called right before a request is sent and right after a response is received, respectively, so something like this api class can add data that is part of every request (setting the header, etc.).  The api class overrides each of these methods

lib/network/src/api.cpp
```diff
@@ -0,0 +1,341 @@
...
+            }
+        } else {
+            request->addPayload("version", API_VERSION);
+            if( authenticated ){
+                request->addParameter("apikey", apiKey);
+            }
+        }
+    }
+
+    void Api::post_send(Request* request, Response* response) {
```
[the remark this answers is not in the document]
mdh5389: authenticated is a flag that is set in the api's public methods.  If it is false, an API key will not be part of the response (authenticated = false).  I wasn't sure if the API key could be sent when registering a scanner so I added this flag.  If I can send an API key (which may or may not be valid) when registering, then I will remove this flag

lib/network/src/api.cpp
```diff
@@ -0,0 +1,341 @@
...
+    int Api::cleanup(Request* req, Response* res){
+        int tmp = res->code;
+        delete req;
+        delete res;
+        return tmp;
+    }
+
+    template<class Type>
+    bool Api::extract(JsonObject& json, Type* data, String key){
+        if(json.containsKey(key)) {
```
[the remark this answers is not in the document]
mdh5389: It only checks on the current level.  The calling method has to make sure it sends the right level.

lib/network/src/api.cpp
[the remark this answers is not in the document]
ThePBJain: not yet, as I haven't done it yet.

lib/core/src/writable.hpp
```diff
@@ -10,10 +10,10 @@ namespace hackPSU{
     public:
         Writable(char* name):  Component(name) { };
         ~Writable() {}
-        virtual void write(Packet* data, uint32_t length = 0){
-            transmit(data, length);
+        virtual void write(Packet data){
```
sushshring: Did the definition for Packet change or are you actually trying to create a new copy for Packet? Should this be a pass by reference?

lib/network/src/api.cpp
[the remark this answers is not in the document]
ThePBJain: the first number should be able to hold data for all the fields in an event. The second should be large enough to hold 25 events.

lib/network/src/api.cpp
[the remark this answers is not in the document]
ThePBJain: nice!

lib/network/src/api.cpp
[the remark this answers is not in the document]
ThePBJain: That works!

lib/network/src/api.cpp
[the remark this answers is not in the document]
ThePBJain: You can pass it, it won't do anything with it but it'll work.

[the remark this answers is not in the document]
ThePBJain: Unless you want it to do something

lib/network/src/api.cpp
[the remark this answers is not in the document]
ThePBJain: Do you want to make an extract that gets the key in any level? That way we only have to ask for the key and it'll find it.

lib/NED/src/esp8266/esp8266.cpp
```diff
@@ -0,0 +1,143 @@
...
+            Serial.println("Start updating " + type);
+        });
+        ArduinoOTA.onEnd([]() {
+            Serial.println("\nEnd");
+        });
+        ArduinoOTA.onProgress([](unsigned int progress, unsigned int total) {
+            Serial.printf("Progress: %u%%\r", (progress / (total / 100)));
+        });
+        ArduinoOTA.onError([](ota_error_t error) {
+            Serial.printf("Error[%u]: ", error);
```
sushshring: Should OTA_enabled be set to false here?

lib/NED/src/esp8266/esp8266.cpp
```diff
@@ -0,0 +1,143 @@
...
+            code = http.GET();
+        } else if(req->isMethod(API::POST)) {
+            String pld = "";
+            req->getPayload().printTo(pld);
+            code = http.POST(pld);
+        }
+    }
+
+    Response* ESP8266_Device::scan() {
+        if( code < 0) {
```
sushshring: nit: mismatching whitespace

lib/NED/src/esp8266/esp8266.cpp
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    void ESP8266_Device::post_send(Request* request, Response* response){
+        Api::post_send(request, response);
+        if(response->payload.length() == 0) {
+            response->code = OUT_OF_MEMORY;
+        } else if(response->code < 0 && status() != WL_CONNECTED){
+            // Resend on a failed attempt
+            connect();
+            if(--retry > 0) {
```
sushshring: nice!

lib/NED/src/esp8266/esp8266.cpp
```diff
@@ -0,0 +1,143 @@
...
+                delete response;
+                write(request);
+                response = read();
+                post_send(request, response);
```
sushshring: Ensure that the recursion here does not cause any issues

lib/network/src/api.cpp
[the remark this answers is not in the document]
mdh5389: That's not what the numbers mean.  The first one is how many JSON Objects are there and the second is how many JSON array elements are there.

lib/core/src/addressable.hpp
```diff
@@ -5,11 +5,23 @@ namespace hackPSU{
...
+    }
+
+    Address getAddress(){
+        return address; 
+    }
+
 private:
     Address address;
+
+    virtual void post_set(){
```
sushshring: If these have no implementation why not make them pure virtual?

lib/network/src/api.cpp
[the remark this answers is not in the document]
mdh5389: Not really.  What if there are multiple entries with the same name (multiple objects in an array each having a UID), then how would we pick which item to extract?

lib/device/src/led.hpp
```diff
@@ -0,0 +1,32 @@
+#include <core.hpp>
+
+#include <Arduino.h>
+
+namespace hackPSU{
+
+    class led: public Writable<int> {
```
sushshring: nit: 
```suggestion
    class Led: public Writable<int> {
```

lib/device/src/led.hpp
```diff
@@ -0,0 +1,32 @@
...
+    public:
+        led(char* name, int pin1): Writable<int>::Writable(name), pin(pin1) {
```
sushshring: nit: name - pin1

lib/core/src/writable.hpp
mdh5389: It changed.  I decided that if you want to make the template object a pointer, you should know how to write that pointer, likewise if it is an object, similar argument.  This way doesn't support writing an array.  That might be a feature to look into

lib/NED/src/esp8266/esp8266.cpp
mdh5389: It shouldn't.  That being said I haven't tested that... Since I delete the response then resend the same request.  It is certainly a good thing to note and I will see what I can do about testing this specific method

lib/core/src/addressable.hpp
mdh5389: Just making it optional.

lib/device/src/led.hpp
mdh5389: Good catch, also this needs to move into directory 'lib/device/src/LED/led.hpp'

lib/network/src/api.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+        Response* res = sendRequest(req, 3);
+        Serial.println("RESPONSE: " + res->payload);
+        if(bool(*res)){
+            MAKE_BUFFER(25, 25) bf_data;
+            JsonObject& response = bf_data.parseObject(res->payload);
+
+            if(response.success()) {
+                JsonObject& data = response.get<JsonObject>("data");
+                if(extract<String>(data, &apiKey, "apikey")) {
```
sushshring: I'm not a huge fan of the three nested if statements here. In my mind this could be done cleaner, but perhaps I'm missing some semantics hidden here.

lib/network/src/api.cpp
```diff
@@ -0,0 +1,341 @@
...
+        return cleanup(req, res);
+    }
+
+    int Api::registerUser(String wid, String pin) {
+        Request* req = createRequest(API::POST, "/rfid/assign");
+        req->addPayload("pin", String(pin));
+        req->addPayload("wid", wid);
+
+        Response* res = sendRequest(req);
+
```
sushshring: nit: whitespace

lib/network/src/api.cpp
```diff
@@ -0,0 +1,341 @@
...
+        if(bool(*res)){
+            MAKE_BUFFER(25, 25) bf_data;
+            JsonObject& response = bf_data.parseObject(res->payload);
+            if(response.success()){
+                if(response.get<int>("length") > 0) {
+                    JsonArray& events = response.get<JsonArray>("locations");
+
+                    for(JsonObject& event : events){
+                        Event* e = new Event();
+                        if(extract(event, e)) {
```
sushshring: chunky nesting

lib/network/src/api.cpp
```diff
@@ -0,0 +1,341 @@
...
+        req->addPayload("location", String(event->id));
+        req->addPayload("timestamp", String(start + offset));
+
+        Response* res = sendRequest(req);
+
+        if(bool(*res)){
+            MAKE_BUFFER(25, 25) bf_data;
+            JsonObject& response = bf_data.parseObject(res->payload);
+
+            if(response.success()){
```
sushshring: These lines seem to be similar in every API request. Is there a specific reason this wasn't made a separate method?

lib/network/src/api.cpp
mdh5389: It's all about the return code.  Outer layer lets us parse the response only if it succeeded.  Inner one makes sure the response is parsable (return API_PARSE_ERROR if not) inner one makes sure all expected items are present (return API_PARSE_ERROR ->should be API_FIELD_MISSING)

lib/network/src/api.cpp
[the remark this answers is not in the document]
ThePBJain: And how do you define JSON objects? The number of key value pairs?

lib/network/src/api.cpp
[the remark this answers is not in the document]
ThePBJain: true.

lib/network/src/api.cpp
mdh5389: Wasn't sure how to dynamically indicate what fields to parse.  It bothers me also

lib/network/src/api.cpp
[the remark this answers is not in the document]
mdh5389: yes

lib/core/src/addressable.hpp
sushshring: ah okay

lib/network/src/api.cpp
sushshring: Perhaps a separate method that does the bf_data, parseObject(res->payload), error checking, and then returns the JsonObject for method specific parsing to the caller.

lib/network/src/api.cpp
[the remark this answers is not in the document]
mdh5389: I think I'll remove the authenticated flag. I'll send the API key on every request

lib/NED/src/esp8266/esp8266.cpp
mdh5389: So this code is copy paste.  I added the concept of OTA_enabled.  In the constructor is it set to false.  At the end of this method it is set to true.

lib/network/src/api.cpp
mdh5389: Have any recommendations to avoid that?

lib/network/src/api.cpp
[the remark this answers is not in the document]
mdh5389: Changed to /scanner

lib/network/src/api.cpp
sushshring: See below

lib/network/src/api.cpp
mdh5389: I think I will do that in a future PR where we are using the newer version of ArduinoJson.  That upgrade will take a bit of work.

bin/codes.sh
```diff
@@ -0,0 +1,66 @@
+# User defined codes
```
david-samuel-mcdermott: Why is this a shell script and not a header file?  HTTP codes don't change and a header file would be more portable

lib/NED/src/esp8266/esp8266.cpp
david-samuel-mcdermott: This will produce something like "Progress: 5%Error[100]"  Because the progress meter lacks a newline, this should have a newline at the beginning.

lib/NED/src/esp8266/esp8266.cpp
david-samuel-mcdermott: Make your life easier, instead of recursing, use a goto here; or set up a do while; knowing this chip, it will come back to get you

lib/NED/src/esp8266/esp8266.hpp
```diff
@@ -0,0 +1,59 @@
...
+    #error Macro, NETWORK_PASSWORD, not set
+    #define NETWORK_PASSWORD "" // Here so intellisense is happy
+#endif
+
+#ifndef NETWORK_SSID
+    #error Macro, NETWORK_SSID, not set
+    #define NETWORK_SSID "" // Here so intellisense is happy
+#endif
+
+#if !defined(OTA_PASSWORD)
```
david-samuel-mcdermott: Why are we using !defined() instead of ifndef?

lib/core/src/addressable.hpp
```diff
@@ -5,11 +5,23 @@ namespace hackPSU{
...
 class Addressable{
 public:
-    Addressable();
-    ~Addressable();
-    void setAddress(Address position) { address = position; }
+    Addressable() {}
+    ~Addressable() {}
+    void setAddress(Address position) { 
+        address = position; 
+        post_set();
```
david-samuel-mcdermott: What is post_set doing, why is it in snake case, and why do we not just anticipate you extending this anyway?

lib/list/src/list.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include <Arduino.h>
+
+namespace hackPSU{
+
+    template <class Item>
+    class List;
+
+    template <class Item>
+    class Wrapper{
```
david-samuel-mcdermott: What is this and why is it named Wrapper; this name is way too generic

lib/list/src/list.hpp
```diff
@@ -0,0 +1,79 @@
...
+            data = (Item*) malloc(size);
+            memcpy(data, item, size);
+        }
+        Item* data;
+        Wrapper<Item>* previous;
+        Wrapper<Item>* next;
+    };
+
+    template <class Item>
+    class List{
```
david-samuel-mcdermott: This needs an implementation and a header file.  I know GCC can be a pain about this, but we need to address this

lib/network/src/api.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+
+    Api::Api(char* name):
+        Network(name)
+    {
+        // Get API key from memory
+        EEPROM.begin(37);
+        loadApiKey();
+
+        #if defined(DEBUG) && defined(SERIAL_EN)
```
david-samuel-mcdermott: We should have a debug print macro if we're doing this.  push the ifs up

lib/network/src/api.hpp
```diff
@@ -0,0 +1,207 @@
...
+#include "codes.h"
+
+#include <list.hpp>
+
+#ifndef API_VERSION
+#error Macro, API_VERSION, not set
+#define API_VERSION ""
+#endif
+
+#define API_PARSE_ERROR -101
```
david-samuel-mcdermott: Do these have meaning beyond being arbitrary numbers, if not, maybe consider enum

bin/codes.sh
mdh5389: I don't believe this file is actually used at the moment and that there is actually a header file for this.  This file should be deleted.

lib/NED/src/esp8266/esp8266.cpp
mdh5389: This is copy paste code.  We can change it but I didn't write any of it

lib/NED/src/esp8266/esp8266.hpp
mdh5389: Because it originally depended on two macros being set (OTA_HASH was the other one)

lib/core/src/addressable.hpp
mdh5389: I don't actually know why for this one

lib/list/src/list.hpp
mdh5389: So you want all the implementation in a cpp file?

lib/list/src/list.hpp
mdh5389: It is used to wrap the templated object so the List class can easily manage it and make it a linked list

lib/network/src/api.cpp
mdh5389: I agree

bin/codes.sh
mdh5389: I deleted that file

lib/list/src/list.hpp
david-samuel-mcdermott: So the C++ standard allows for it, it's just a matter of how shit our compiler is, but if it doesn't want to, then don't worry about it; we'll spend as much time figuring that out as we would an actual problem

lib/network/src/api.cpp
david-samuel-mcdermott: I think we _should_ do it here, it'll just be something like

```
#if defined(DEBUG) and defined(SERIAL_EN)
#define debug_printf(format, ...) printf(format, __VA_ARGS__)
#else 
#define debug_printf(format,...)
#endif
```

lib/network/src/api.hpp
david-samuel-mcdermott: Gotcha, maybe add a comment explaining that, so that we know these don't strictly mean anything other than our own labels; the decision to go with -101 seems arbitrary, and it will make onboarding easier

mdh5389: See update.

lib/network/src/api.cpp
mdh5389: Where should the be defined though?

jmm8046:
If you think it's good to PR in then we can get it in, has it been tested?

david-samuel-mcdermott:
90% sure this is what we used for the hackathon, but try it, I'm forwarding y'all build instructions

sushshring:
I think it just never got reviewed/had some pending minor fixes. Someone should go through and make sure it is coherent with whatever redis might be doing (if it is doing) before deploying this to the actual event

sushshring:
👀 52 commits though :P

david-samuel-mcdermott:
This happened  because it was just @mdh5389 on firmware for a long while.  Also @sushshring I approved bc master won't build, so YOLO

PR #33, opened 2019-10-14
Create PoleMaster_0.1.py

mazlow17:
PoleMasterCode Beta Progression

### DIFF
--- a/PoleScanner_0.1/PoleMaster_0.1.py
+++ b/PoleScanner_0.1/PoleMaster_0.1.py
@@ -18,6 +18,7 @@ import threading
 ###LOGIC IMPORT###
 import urllib2
 import sys
+import os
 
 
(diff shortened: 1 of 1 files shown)


jmm8046:
Are we sure we want to combine this onto the scanner repo and not have it as it's own repo? Open to thoughts for either choice

david-samuel-mcdermott:
@jmm8046 You want two repos - if these codebases don't touch, you want two separate repos unless you have a good reason to have them in one - it will get confusing and it lets you hint where people can work easier (this is like 90% C++ the other repo would be like 90% Python)

jmm8046:
@david-samuel-mcdermott Yep! We already made a different repo and just forgot to close this PR

david-samuel-mcdermott:
Okay, let's do that then

